Issue #2, opened 2020-09-10
Scope question: index creation


slazicoicr:
The crates currently offers read and write functionality for various file formats and their respective index formats. Is it within the scope of this repo to include an `Indexer` struct to calculate the index `Record` from the respecitve file format?

zaeleus:
Hi @slazicoicr!

Yes, index builders are on the todo list, particularly for BAM.

zaeleus:
As of 221bb0980f971ec3b75e867568ef82a6d1bafa9e, BAM, CRAM, FASTA, and FASTQ have indexers. GFF, SAM, and VCF can be indexed via tabix. If there are any further problems with indexing, please submit a new issue.

Thanks again to @slazicoicr for working on the FASTA indexer (#4)!

PR #3, opened 2020-09-12
Added `read_sequence_raw` function to fasta reader

slazicoicr:
I would like to implement the `fasta` index builder (discussion about index builders at #2): http://www.htslib.org/doc/faidx.html . The index requires line length information, which is currently discarded by `read_sequence` (the newlines are trimmer). The new function retains the newlines, allowing for inference of line length.

There is a lot of code repetition between `read_sequence` and `read_sequence_raw`, but I am unsure if switching `read_sequence` to filter out the newlines from a `read_sequence_raw` call would affect performance.

### DIFF
--- a/noodles-fasta/src/reader.rs
+++ b/noodles-fasta/src/reader.rs
@@ -73,8 +73,8 @@ where
     /// The position of the stream is expected to be at the start of a sequence, which is directly
     /// after a definition.
     ///
-    /// If successful, this returns the number of bytes read from the stream. If the number of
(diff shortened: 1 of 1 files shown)


zaeleus:
I understand capturing the newlines in the sequence is necessary to build a FASTA index, but I don't think it has any other use cases? I expect a different reader would have to be used to build the index, as you shouldn't need to read the entire sequence into memory.

slazicoicr:
You are right. I will actually prototype the index builder rather than try to predict what will be needed.

PR #4, opened 2020-09-16
Fasta index builder

slazicoicr:
Would like input for

* Naming of structs/methods
* Module organization
* How indices should be constructed: `Vec<Record>` or an iterator that returns `Records`

### DIFF
--- a/noodles-fasta/src/fai.rs
+++ b/noodles-fasta/src/fai.rs
@@ -5,4 +5,4 @@ mod reader;
 mod record;
 mod writer;
 
-pub use self::{reader::Reader, record::Record, writer::Writer};
+pub use self::{indexer::index, reader::Reader, record::Record, writer::Writer};
--- a/noodles-fasta/src/fai.rs
+++ b/noodles-fasta/src/fai.rs
@@ -1,8 +1,8 @@
 //! FASTA index (FAI) and fields.
 
+mod builder;
 mod reader;
 mod record;
 mod writer;
-mod builder;
 
-pub use self::{reader::Reader, record::Record, writer::Writer, builder::IndexBuilder};
+pub use self::{builder::IndexBuilder, reader::Reader, record::Record, writer::Writer};
--- a/noodles-fasta/src/fai.rs
+++ b/noodles-fasta/src/fai.rs
@@ -5,4 +5,4 @@ mod reader;
 mod record;
 mod writer;
 
-pub use self::{indexer::Indexer, reader::Reader, record::Record, writer::Writer};
+pub use self::{reader::Reader, record::Record, writer::Writer};
--- a/noodles-fasta/src/fai.rs
+++ b/noodles-fasta/src/fai.rs
@@ -3,5 +3,6 @@
 mod reader;
 mod record;
 mod writer;
+mod builder;
 
-pub use self::{reader::Reader, record::Record, writer::Writer};
+pub use self::{reader::Reader, record::Record, writer::Writer, builder::IndexBuilder};
--- a/noodles-fasta/src/fai.rs
+++ b/noodles-fasta/src/fai.rs
@@ -1,8 +1,8 @@
 //! FASTA index (FAI) and fields.
 
-mod builder;
+mod indexer;
 mod reader;
 mod record;
 mod writer;
 
-pub use self::{builder::IndexBuilder, reader::Reader, record::Record, writer::Writer};
+pub use self::{indexer::Indexer, reader::Reader, record::Record, writer::Writer};
--- a/noodles-fasta/src/fai/indexer.rs
+++ b/noodles-fasta/src/fai/indexer.rs
@@ -1,6 +1,13 @@
 //! Index creation for a FASTA file as specified at http://www.htslib.org/doc/faidx.html
 
-use std::{error::Error, fmt, io, io::BufRead};
+use std::{
+    error::Error,
+    fmt,
+    fs::File,
+    io,
+    io::{BufRead, BufReader},
+    path::Path,
+};
 
 use memchr::memchr;
 
@@ -13,30 +20,29 @@ use super::Record;
 
 /// Creates an index from a FASTA file.
 ///
-/// The position of the stream is expected to be at the beginning of the file.
-///
 /// # Examples
-/// ```
-/// use noodles_fasta::fai::{index, Record};
-/// let data = b"\
-/// >one
-/// ATGCAT
-/// GCATGC
-/// ATG
-/// >two another chromosome
-/// ATGC
-/// GCAT";
 ///
-/// let indx = index(&data[..]).expect("Failed to read or invalid format");
-/// let expected = vec![
-///     Record::new("one".to_string(), 15, 5, 6, 7),
-///     Record::new("two".to_string(), 8, 47, 4, 5)
-/// ];
+/// Write a FASTA index file to disk.
 ///
-/// assert_eq!(indx, expected);
+/// ```no_run
+/// use std::{fs::File, io::Result};
+/// use noodles_fasta::fai::{index, Writer};
+///
+/// fn main() -> Result<()> {
+///     let mut fai_writer = File::create("foo.fai").map(Writer::new)?;
+///
+///     for record in index("foo.fa")? {
+///         fai_writer.write_record(&record)?;
+///     }
+///     Ok(())
+/// }
+///
 /// ```
-pub fn index<R: BufRead>(buf: R) -> Result<Vec<Record>, IndexError> {
-    let mut indexer = Indexer::new(buf);
+pub fn index<P>(src: P) -> io::Result<Vec<Record>>
+where
+    P: AsRef<Path>,
+{
+    let mut indexer = File::open(src).map(BufReader::new).map(Indexer::new)?;
     let mut result = Vec::new();
 
     while let Some(i) = indexer.index_record()? {
@@ -100,7 +106,7 @@ where
     }
 
     /// Indexes a raw FASTA record.
-    /// 
+    ///
     /// The position of the stream is expected to be at the start or at the start of another
     /// definition.
     ///
@@ -147,7 +153,7 @@ where
         }
 
         if length == 0 {
-            return Err(IndexError::EmptyDefinition(self.offset));
+            return Err(IndexError::EmptySequence(self.offset));
         }
 
         let record = Record::new(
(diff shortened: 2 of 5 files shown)


zaeleus (changes requested):
Currently, the index is simply a list of index records, i.e., `Vec<fai::Record>`. Ultimately, this is what should be returned, so we should focus on the public API being as simple as `pub fn index<P>(src: P) -> io::Result<Vec<fai::Record>> where P: AsRef<Path>`. This should be the only function that needs to be exposed at this point.

slazicoicr:
All changes are in. Public API suggestion makes sense and I can attempt them in a separate PR.

noodles-fasta/src/fai.rs
```diff
@@ -1,7 +1,8 @@
 //! FASTA index (FAI) and fields.
 
+mod indexer;
 mod reader;
 mod record;
 mod writer;
 
-pub use self::{reader::Reader, record::Record, writer::Writer};
+pub use self::{indexer::Indexer, reader::Reader, record::Record, writer::Writer};
```
zaeleus, line 8:
```suggestion
pub use self::{reader::Reader, record::Record, writer::Writer};
```

As previously mentioned, `Indexer` should remain private.

noodles-fasta/src/fai/indexer.rs
```diff
@@ -0,0 +1,240 @@
...
+            first_index,
+            Some(Record::new("one".to_string(), 66, 6, 30, 32))
+        );
+        let second_index = builder.index_record().unwrap();
+        assert_eq!(
+            second_index,
+            Some(Record::new("two".to_string(), 28, 103, 14, 16))
+        );
+        assert_eq!(None, builder.index_record().unwrap());
+    }
```
zaeleus, line 239: This can be replaced with a unit test for `len_with_right_trim`.

noodles-fasta/src/fai/indexer.rs
```diff
@@ -0,0 +1,240 @@
...
+        if def_len == 0 {
+            return Ok(None);
+        }
+
+        self.offset += def_len;
+        let index_offset = self.offset;
+        let def: Definition = def_str.trim_end().parse()?;
+
+        // The first sequence line determines how long each line should be
+        let mut seq_len = self.read_sequence_line(&mut line)?.try_into().unwrap();
```
zaeleus, line 115: Please revert this (and others) to be consistent with the rest of the library.

noodles-fasta/src/fai/indexer.rs
```diff
@@ -0,0 +1,240 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_read_sequence_line() {
+        let data = b"ACGT\nNNNN";
```
zaeleus, line 208:
```suggestion
        let data = b"ACGT\nNNNN\n";
```

noodles-fasta/src/fai/indexer.rs
```diff
@@ -0,0 +1,240 @@
+//! Creates a index for a FASTA file
+
+use std::convert::TryInto;
+use std::error::Error;
+use std::fmt;
+use std::io;
```
zaeleus, line 6: Merge imports.

noodles-fasta/src/fai/indexer.rs
```diff
@@ -0,0 +1,240 @@
...
+
+use super::Record;
+use crate::reader::{DEFINITION_PREFIX, NEWLINE};
+use crate::record::definition::{Definition, ParseError};
+
+use memchr::memchr;
```
zaeleus, line 12: Separate imports into groups of std -> external -> crate -> super.

noodles-fasta/src/fai/indexer.rs
```diff
@@ -0,0 +1,240 @@
...
+
+/// A index builder
+pub struct Indexer<R> {
+    inner: R,
+    offset: u64,
+}
+
+impl<R> Indexer<R>
+where
+    R: io::BufRead,
```
zaeleus, line 22:
```suggestion
    R: BufRead,
```

Import `std::io::BufRead`.

noodles-fasta/src/fai/record.rs
```diff
@@ -8,7 +8,7 @@ const FIELD_DELIMITER: char = '\t';
 const MAX_FIELDS: usize = 5;
 
 /// A FASTA index record.
-#[derive(Debug, Default)]
+#[derive(Debug, Default, PartialEq, Eq)]
```
zaeleus, line 11:
```suggestion
#[derive(Debug, Default, Eq, PartialEq)]
```

zaeleus (changes requested):
* Add punctuation to short descriptions and comments.
* Avoid `unwrap` in examples and tests.
* Please include a public function in this changeset.

slazicoicr:
All changes are in. The public function signature is `pub fn index<R: BufRead>(buf: R) -> Result<Vec<Record>, IndexError>`

zaeleus:
> `pub fn index<R: BufRead>(buf: R) -> Result<Vec<Record>, IndexError>`

This isn't quite the signature I envisioned previously.

Overall, the changes are okay for now. Instead of dragging this on, I can merge it as is and make stylization changes. Is this something you prefer, or do you want to do it via review suggestions?

slazicoicr:
I am happy to keep dinging at it. 

The suggestion from early on was `pub fn index<P>(src: P) -> io::Result<Vec<fai::Record>> where P: AsRef<Path>`, but I wasn't sure if that call should also be able to deal with `fasta.gz` files. The current signature can be turned private, as it can be used to deal with streams and in memory fasta sequences.

noodles-fasta/src/fai/indexer.rs
```diff
@@ -0,0 +1,247 @@
...
+            if seq_len == 0 {
+                break;
+            // If there are more lines, check the previous line has equal length to first.
+            } else if prev_line_width != line_width || prev_line_bases != line_bases {
+                return Err(IndexError::InvalidLineLength(self.offset));
+            }
+        }
+
+        if length == 0 {
+            return Err(IndexError::EmptyDefinition(self.offset));
```
zaeleus, line 142: If `length` is 0, doesn't that mean the sequence is empty and not the definition?

noodles-fasta/src/fai/indexer.rs
```diff
@@ -0,0 +1,247 @@
+//! Index creation for a FASTA file as specified at http://www.htslib.org/doc/faidx.html
+
+use std::{error::Error, fmt, io, io::BufRead};
+
+use memchr::memchr;
+
+use crate::reader::{DEFINITION_PREFIX, NEWLINE};
+use crate::record::definition::{Definition, ParseError};
```
zaeleus, line 8:
```suggestion
use crate::{
    reader::{DEFINITION_PREFIX, NEWLINE},
    record::definition::{Definition, ParseError},
};
```

noodles-fasta/src/fai/indexer.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    /// Reads a single sequence line.
+    ///
+    /// Trailing whitespaces are not modified.
+    ///
+    /// Returns the number of bytes read from the stream. If the position of the stream is at
+    /// a definition or EOF has been reached, the number of bytes read is 0.
+    fn read_sequence_line(&mut self, buf: &mut Vec<u8>) -> io::Result<usize> {
+        let mut bytes_read = 0;
+        let mut end_line = false;
```
zaeleus, line 70:
```suggestion
        let mut is_eol = false;
```

noodles-fasta/src/fai/indexer.rs
```diff
@@ -0,0 +1,247 @@
...
+
+        loop {
+            let reader_buf = self.inner.fill_buf()?;
+
+            if end_line || reader_buf.is_empty() || reader_buf[0] == DEFINITION_PREFIX {
```
zaeleus, line 75:
```suggestion
            if is_eol || reader_buf.is_empty() || reader_buf[0] == DEFINITION_PREFIX {
```

noodles-fasta/src/fai/indexer.rs
```diff
@@ -0,0 +1,247 @@
...
+                break;
+            }
+
+            let len = match memchr(NEWLINE, reader_buf) {
+                Some(i) => {
+                    // Do not consume newline.
+                    buf.extend(&reader_buf[..=i]);
+                    end_line = true;
```
zaeleus, line 83:
```suggestion
                    is_eol = true;
```

zaeleus (changes requested):
Sorry for the large suggestion blocks; it seems GitHub doesn't show proper diffs for them. The changes are mainly stylization and wording.

zaeleus:
> I wasn't sure if that call should also be able to deal with `fasta.gz` files.

Indexing a compressed FASTA isn't useful without an associated bgzip index (gzi), which is out of scope for this patch.

slazicoicr:
Question about `pub fn index<P>(src: P) -> io::Result<Vec<fai::Record>> where P: AsRef<Path>` testing approach. Would it be necessary to include a (small) .fasta file to load during doc string testing (if so, where is the spot for it)?

Another approach is to keep (and rename) `fn index_from_bufread<R: BufRead>(buf: R) -> Result<Vec<Record>, IndexError>` (no longer public) and have tests done on it. The public file path `index` function would just be a simple extension of `index_from_bufread`.

The last option is not to do any testing with `pub fn index<P>(src: P) -> io::Result<Vec<fai::Record>> where P: AsRef<Path>`, as it is fairly simple function.

zaeleus:
You can add an example to test compilation and not run it. This can be done with the [`no_run` attribute](https://doc.rust-lang.org/rustdoc/documentation-tests.html#attributes) on a code block. For example, https://github.com/zaeleus/noodles/blob/0cbea550aa5cf2710a92fa947f428de07fd0fa3d/noodles-fasta/src/lib.rs#L33-L53

slazicoicr:
Public `index` function now takes a file path.

zaeleus:
Nice, thanks!

slazicoicr:
Anything left for this PR?

zaeleus (approved):

slazicoicr:
Don't have permissions to merge.

zaeleus:
Yes, please have patience, @slazicoicr; it will be merged. :)

PR #7, opened 2020-10-18
Tutorial and/or Structopt for parsing input rather than .nth(1)

otsukaresamadeshita:
Hi Zaeleus, thanks for the awesome package! After struggling with rust::htslib a little, I found your package and it is a large undertaking to say the least. I would be happy to contribute in any way I can, but am still learning the basics. In any case, I was interested in using noodles and thought I would do a writeup of how I basically used it. It can be found in the noodles-tutorial folder. When I was trying to write some code with the mentioned noodles-bam = { git = "https://github.com/zaeleus/noodles.git" } I got back errors that rust couldn't find some items that it needed etc etc so I think this will help.

I also was thinking of replacing this in each of the examples with StructOpt from the rust tutorial structopt = "0.3.20"
let src = env::args().nth(1).expect("missing src");

I'm hoping to use noodles in my future projects, Cheers!

Now passes CI, I made a git mistake, but should have the short tutorial now.


zaeleus:
Hi, @otsukaresamadeshita!

Unfortunately, argument parsing is out of the scope of the documentation and examples included in the library. Your tutorial may be better suited as a personal blog post or perhaps shared on [users.rust-lang.org](https://users.rust-lang.org/) or [/r/rust](https://www.reddit.com/r/rust/). Thanks, though!

> When I was trying to write some code with the mentioned noodles-bam = { git = "https://github.com/zaeleus/noodles.git" } I got back errors that rust couldn't find some items that it needed

Please submit an issue if this is the case. git repositories are [supported by Cargo][0], and there is no need to explicitly clone them.

[0]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

Issue #11, opened 2020-12-24
Buffered reading of BGZF breaks calculation of virtual positions.


MaltheSR:
While experimenting, I ran into a quirk in the `bgzf` module: when working with a `reader: io::BufReader<bgzf::Reader<R>>`, calls to `reader.get_ref().virtual_position()` break. A little MWE to illustrate:

```
use std::io::{self, BufRead, Read, Write};

use noodles_bgzf as bgzf;

fn log(line: &str, virtual_position: &bgzf::VirtualPosition) {
    println!(
        "Line '{}', compressed offset {}; uncompressed offset {}.",
        line, virtual_position.compressed(), virtual_position.uncompressed()
    );
}

fn main() -> io::Result<()> {
    // Setup

    let input = b"line1\nline2\nline3\nline4\nline5\n";

    let mut writer = bgzf::Writer::new(Vec::new());
    writer.write_all(input)?;
    let data = writer.finish()?;

    // Plain reader (this works as expected)

    println!("Plain reader:");

    let mut reader = bgzf::Reader::new(data.as_slice());
    let mut virtual_position = reader.virtual_position();
    let mut buf = [0u8; 6];

    while let Ok(_) = reader.read_exact(&mut buf) {
        let line = std::str::from_utf8(&buf[0..5]).expect("cannot parse UTF8");
        log(line, &virtual_position);
        virtual_position = reader.virtual_position();
    };

    println!();

    // Buffered reader (this does not work as expected)

    println!("Buffered reader:");

    let mut reader = io::BufReader::new(bgzf::Reader::new(data.as_slice()));

    loop {
        let virtual_position = reader.get_ref().virtual_position();
        let mut line = String::new();
        match reader.read_line(&mut line) {
            Ok(0) => break,
            Err(e) => return Err(e),
            _ => { line.pop(); },
        }
        log(&line, &virtual_position);
    }

    Ok(())
}
```

Running this prints:

```
Plain reader:
Line 'line1', compressed offset 0; uncompressed offset 0.
Line 'line2', compressed offset 0; uncompressed offset 6.
Line 'line3', compressed offset 0; uncompressed offset 12.
Line 'line4', compressed offset 0; uncompressed offset 18.
Line 'line5', compressed offset 0; uncompressed offset 24.

Buffered reader:
Line 'line1', compressed offset 0; uncompressed offset 0.
Line 'line2', compressed offset 46; uncompressed offset 0.
[...]

PR #12, opened 2020-12-26
bgzf: impl BufRead for Reader

MaltheSR:
I sketched a draft solution to #11 out of curiosity, and I figured the issue might be more constructive with an illustration, so I'll go ahead and make it a PR if you're interested. Feel free to completely disregard this if you don't like it. 

By refactoring `Block` to keep track of both `upos` and `cpos` on a raw `Vec<u8>` (instead of a `Cursor`), the `Reader` now implements `BufRead` in a similar manner to a `BufReader`, except the `upos` can be correctly tracked in the `consume` call. This allows the motivating case in #11 to work, as demonstrated by a new test in `noodles-bgzf/src/lib.rs`

This would still need (at least) more testing, updates on module docs, and some changes downstream where a `BufReader<bgzf::Reader<R>>` is used, but I'm leaving that outside of this MWE.


zaeleus:
@MaltheSR, I'm staging this to rebase and merge unless you have any other outstanding issues or concerns.

The [current clippy failure](https://github.com/zaeleus/noodles/runs/1631675314) is irrelevant to the patch but fixed in ca70e7416127bd69fcd572cac4501e39b14484ec.

MaltheSR:
@zaeleus, thank you very much for taking the time to look at this, and apologies that the initial PR was a bit of a lump commit. Good catch on BGZF block ISIZEs not fitting in a u16, I hadn't realised that was off-by-one.

After merging this commit, I figure that `io::BufReader<bgzf::Reader<R>>` will be an anti-pattern, so I grepped through the rest of noodles to catch instances of this. The above three commits simply change `io::BufReader<bgzf::Reader<R>>` to `bgzf::Reader<R>>` to rely on the new `bgzf::Reader::BufRead` directly.

Feel free to merge as you see fit, and thanks again!

Issue #15, opened 2021-05-13
Accessing (potentially missing) tags and reading values


jamestwebber:
Hey there! I've got a BAM-parsing pipeline that I'd love to re-implement in Rust using `noodles` to speed everything up and consolidate many steps.

I've been reading through the code/documentation/examples, which are all great! But I either haven't found or haven't grokked how to a) check for the presence of a tag and b) get the value in an efficient way.

Using `pysam` the workflow would be something like (for example):

```python
for a in aligned_bam:
   if a.is_secondary or a.is_supplementary or a.mapping_quality < mq:
       continue
   for tag in tags:  # I have a list of tags I'm looking for
       if a.has_tag(tag):
           tag_histograms[tag][a.get_tag(tag)] += 1
```

I'm not sure how I would do this with `noodles`. My current guess is

```rust
for result in reader.records() {
    let record = result?;        
    let flags = record.flags();
    let data = record.data();

    if !(flags.is_supplementary() || flags.is_duplicate()) {
        let mapq = u8::from(record.mapping_quality());

        if mapq != 255 && mapq >= 10 {
            // iterate over data and add values to counters?
        }
    }
}
```

In part I'm questioning whether I need to iterate over the tags or if there's a faster lookup method. That's about where I decided to open an issue.

Happy to contribute examples for this workflow, if you can point me in the right direction!

zaeleus:
Hi, @jamestwebber!

`bam::record::Data::fields` returns an iterator over the data fields (effectively, `Iterator<Item = io::Result<bam::record::data::Field>>`). This allows an end user to choose the most efficient lookup method for their use case. For example, a linear search would perform better than creating an entire map for a single tag lookup.

In your case, I would suggest creating and using a map for multiple lookups. You can either create one with the BAM data fields

```rust
let data: HashMap<_, _> = record
    .data()
    .fields()
    .map(|result| result.map(|field| (field.tag().clone(), field))) 
    .collect::<Result<_, _>>()?;
```

or convert it to SAM data, which is stored as a tag-field map.

```rust
let data = sam::record::Data::try_from(bam_record.data())?;
```

With this, you can

> a) check for the presence of a tag

```rust
data.contains_key(&Tag::AlignmentScore);
```

> b) get the value in an efficient way

```rust
let value = data.get(&Tag::AlignmentScore).map(|field| field.value());
```

The Rust code using noodles then becomes similar to your Python example.

```rust
let tags = [
    Tag::MinMappingQuality,
[...]

PR #16, opened 2021-05-17
Implementing DerefMut for sam::record::Data

jamestwebber:
Now that I've got the code from #15 working it really opened the floodgates 😁 

My next target: reading a BAM file, filtering by checking for a specific tag, and then writing a slightly modified BAM file with a new tag added. Specifically I'm checking for a barcode tag that's part of my selected set, and then adding a corrected barcode in a new tag and writing it to my new BAM file.

Unfortunately I got stuck on the last step. I could create a new `sam::Record` with the additional tag, but it felt clumsy to me--it didn't appear that there was a natural way to copy things, so I would need to use `Record::builder` and manually set all the fields (including checking for errors).

I thought maybe I could use the Rust structure-copying method of `Record{ data: new_data, ..old_record }` but all of the fields are private so it won't let me do that.

One way that I saw that would make this work is if I could borrow a mutable pointer to the Record's data, e.g. 

```
let mut sam_record = record.try_into_sam_record(header.reference_sequences())?;
let sam_data = sam_record.data_mut();
sam_data.insert(my_new_tag, my_new_field);
writer.write_sam_record(header.reference_sequences(), &sam_record)?;
```

I needed to implement `DerefMut` for the Data struct and add `data_mut` to the Record struct. That's what this PR does, and it works!

You might prefer to keep everything immutable instead of allowing this, which I can understand. In that case I'd request a more convenient way to modify an existing record, as (at least for me) modifying a record with some new data is a pretty common use-case.

### DIFF
--- a/noodles-sam/src/record.rs
+++ b/noodles-sam/src/record.rs
@@ -342,26 +342,27 @@ impl Record {
         &self.data
     }
 
-    /// Returns a mutable reference to the data fields
+    /// Returns a mutable reference to the data fields for this record.
     ///
     /// # Examples
     ///
     /// ```
-    /// use noodles_sam::{self as sam, record::{data, Data}};
+    /// use noodles_sam::{self as sam, record::data};
     ///
     /// let mut record = sam::Record::default();
     /// assert!(record.data().is_empty());
     ///
-    /// let data = record.data_mut();
-    /// data.insert(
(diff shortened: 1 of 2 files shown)


zaeleus:
Nice, thanks for the descriptive use case and patch!

PR #22, opened 2021-06-21
Make `seek_to_first_record` public

jamestwebber:
Tiny tweak that I think is worth making: the method `seek_to_first_record` would be nice to have available as a public method as it removes a little boilerplate for simple scripts.

In my case I need to read through the BAM twice so it's handy to have this available (and I don't need to bring `VirtualPosition` into my module).

### DIFF
--- a/noodles-bam/src/reader.rs
+++ b/noodles-bam/src/reader.rs
@@ -262,7 +262,7 @@ where
 
     // Seeks to the first record by setting the cursor to the beginning of the stream and
     // (re)reading the header and binary reference sequences.
-    fn seek_to_first_record(&mut self) -> io::Result<VirtualPosition> {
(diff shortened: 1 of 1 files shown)


zaeleus:
I'm hesitant to make this public, as the visibility change is consequential to other `Reader` interfaces in terms of consistency. I understand the convenience, but for now, I'll have to decline, sorry.

> and I don't need to bring `VirtualPosition` into my module

If you _really_ don't want to import the type, you can call the default trait method for the seek.

```rust
fn seek_to_first_record(reader: &mut bam::Reader) -> io::Result<()> {
    reader.seek(Default::default())?;
    reader.read_header()?;
    reader.read_reference_sequences()?;
    Ok(())
}
```

Issue #29, opened 2021-07-12
Publish a version of things on crates.io


sstadick:
Is there a timeline or line in the sand for when you might be interested in pushing a v0.1.0 of all of these to `crates.io`? It's impossible to publish a crate that depends on any of the `noodles` family until there is a version of them in `crates.io`. 

If there's a checklist of things to add to add or do for an initial release I'm happy to help push it over the finish line. Specifically I want to release a new version of [perbase](https://github.com/sstadick/perbase/pull/38) that uses `noodles-bgzf` can't publish it to crates.io until until noodles is there first.

brainstorm:
For now my GSoC 2021 students have been mostly using HEAD, but I agree some release model needs to happen sooner or later. 

I hope that PR #30 bootstraps this so we can have noodles over `crates.io` soon enough ;)

zaeleus:
I'll look into publishing a release this week.

zaeleus:
noodles is now published on crates.io: https://crates.io/crates/noodles.

sstadick:
That's fantastic! Thank you!

PR #30, opened 2021-07-13
Release deployment

brainstorm:
Inspired by the commits I've seen over https://github.com/rust-bio/rust-htslib/blob/master/.github/workflows/release-please.yml.

Ideally we should be using https://github.com/googleapis/release-please/pull/954 as a way to easily publish all sub-crates within this repo, instead of enumerating each one.

Fixes issue #29 

/cc @sstadick

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,40 @@
+on:
+  push:
+    branches:
+      - master
+
+name: CD 
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: GoogleCloudPlatform/release-please-action@v2
+        id: release-noodles 
+        with:
+          release-type: rust
+          package-name: noodles 
+          bump-minor-pre-major: Yes
+          monorepo-tags: true
+
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
+        if: ${{ steps.release_main.outputs.release_created || steps.release_sys.outputs.release_created }}
+        with:
+          toolchain: stable
+          override: true
+
+      - name: Install system dependencies
+        if: ${{ steps.release_main.outputs.release_created || steps.release_sys.outputs.release_created }}
+        run: |
+            sudo apt-get install --yes zlib1g-dev libbz2-dev musl musl-dev musl-tools clang libc6-dev build-essential
+
+      - uses: Swatinem/rust-cache@v1.3.0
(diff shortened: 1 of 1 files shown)


brainstorm:
Actually, that release-please PR is just a wrapper around https://github.com/fasterthanlime/cargo-workspaces#publish... so I guess we could just use that in the meantime to publish all crates within the noodles workspace.

sstadick:
An alternative method: https://github.com/BurntSushi/ripgrep/blob/master/RELEASE-CHECKLIST.md where something like `cargo workspace publish` is just run locally.

In the yaml I am personally a fan of only triggering on tags of some sort https://github.com/sstadick/perbase/blob/030904f6178f3fd4c10fcc97379d072abb0b39e4/.github/workflows/main.yml#L5, or `v.*` as the pattern.

brainstorm:
Thanks @sstadick! By the comment on #29 I'd guess that @zaeleus might publish the first version manually this week before automating it? In any case, let's see which option he prefers from the ones described here.

sstadick:
I'm happy regardless of the method a release happens, and super excited to see a version of this on `crates.io`!

zaeleus:
Thanks, but I'd like to publish manually for now, particularly to better understand the process when managing many interdependent crates. For example, there's a small delay between a successful publication and when you can actually use the crate, which can make an automated publication of a crate that depends on it fail. I also learned that crates.io enforces a rate limit after 5 pushes. :P

```
error: failed to publish to registry at https://crates.io

Caused by:
  the remote server responded with an error (status 429 Too Many Requests): You
  have published too many crates in a short period of time. Please try again
  after Wed, 14 Jul 2021 22:34:44 GMT or email help@crates.io to have your limit
  increased.
```

brainstorm:
Yeah, I understand, those limitations/gotchas on `crates.io` are indeed unexpected.

I wonder how @fasterthanlime deals with that kind of rate limiting and dependency (timing) failures when using the recently merged https://github.com/googleapis/release-please/pull/954.

fasterthanlime:
We use "cargo workspaces publish", which polls the registry after publishing to wait for it to really show up.

We publish to a private Cloudsmith registry so we haven't run into the crates.io rate limiting!

zaeleus:
Just for future reference, I talked to a crates.io team member. He clarified that there is a 5 _new_ crate burst publication rate limit and a 30 _existing_ crate burst publication rate limit (1 publication/minute).

Issue #17, opened 2021-05-17
parallelization of bgzf compression


KHanghoj:
hi,

I used noodles to modify and filter bam files with the primary focus of learning rust. Is it possible to parallelize the compression part when writing to a bam file. 

Great package :)

Kristian

jamestwebber:
Hah! I was just thinking about this as well. Unfortunately I think it would be rather complicated to implement--you'd want to parallelize the compression step but preserve the original ordering when writing out the records. It would be super cool though!

KHanghoj:
ha.

Given my current rust skills im not of much help but some of the multi-threading in htslib can be found here: https://github.com/samtools/htslib/blob/515f6df8f7f7dab6c80d0e7aede6e60826ef5374/bgzf.c#L1722 i believe. 

Dont know if that is of any help.

This is relevant for BAM/bcf format ( i think) as the threads are only used form compression

jamestwebber:
Note: I'm not @zaeleus so I can't speak for their plans, but I just started implementing some things using the library and parallel writes were on my mind.

I'm a little confused about whether the htslib implementation preserves read order. It looks like they have a threadpool to do the compression and they push the results to a queue, and it seems like that might reorder the reads slightly. Maybe that's accepted behavior, I couldn't find any documentation on this.

I was trying to find a way to preserve ordering but it's a tricky problem, e.g. see [this Rayon issue](https://github.com/rayon-rs/rayon/issues/210) which has been open for quite a while. But I think the rayon folks are looking for an extremely general solution, while the heap-queue thing proposed in that issue would probably work great for BAM processing (maybe with a little parameter tweaking).

I'm not sure I'm gonna try it, but if you want to try implementing it as a way to learn a ton of Rust all at once, I'd love to see it. 😁

jkbonfield:
The htslib thread pool is complicated!

The pool has the notion of input (things to do) and output (completed jobs) queues.  The output queue can be read from either as they arrive or in strict order of submission.  Clearly with BGZF it is critical to get the ordering correct, so that's how it utilises the thread pool.

If we're reading and writing data (eg BAM to BAM, BAM to CRAM, etc) then we have one shared thread pool with multiple input and output queues attached to cover the decode and encode tasks.  The tricky bit was stopping deadlocks when a queue fills up, while limiting the number of items (decoding is vastly faster than encoding so you don't want to buffer up most of the uncompressed file because it's running ahead of the encoder tasks).

[...]

PR #36, opened 2021-08-14
builder methods should take ownership

jamestwebber:
I was excited to try out the new multi-threaded async functions, but I ran into what I think is a bug. If I try to build an AsyncWriter using the Builder and `set_worker_count`, it will fail to compile with

```
error[E0507]: cannot move out of a mutable reference
  --> src/retag_bam.rs:54:22
   |
54 |       let mut writer = bam::AsyncWriter::builder(file)
   |  ______________________^
55 | |         .set_worker_count(8)
56 | |         .set_compression_level(Compression::fast())
   | |___________________________________________________^ move occurs because value has type `bam::r#async::writer::builder::Builder<tokio::fs::File>`, which does not implement the `Copy` trait

error: aborting due to previous error
```

I believe this is because the builder methods were by reference, and they need to take ownership to return the modified object.

I only modified the bam::AsyncWriter::Builder object here but if you think this is the right change I am happy to make the same tweak for other Builders.

### DIFF
--- a/noodles-bam/src/async/reader/builder.rs
+++ b/noodles-bam/src/async/reader/builder.rs
@@ -31,7 +31,7 @@ where
     /// let data = [];
     /// let builder = bam::AsyncReader::builder(&data[..]).set_worker_count(8);
     /// ```
-    pub fn set_worker_count(&mut self, worker_count: usize) -> &mut Self {
+    pub fn set_worker_count(mut self, worker_count: usize) -> Self {
         self.worker_count = Some(worker_count);
         self
     }
--- a/noodles-bam/src/async/writer/builder.rs
+++ b/noodles-bam/src/async/writer/builder.rs
@@ -36,7 +36,7 @@ where
     /// let builder = bam::AsyncWriter::builder(Vec::new())
     ///     .set_compression_level(Compression::best());
     /// ```
-    pub fn set_compression_level(&mut self, compression_level: Compression) -> &mut Self {
+    pub fn set_compression_level(mut self, compression_level: Compression) -> Self {
         self.compression_level = Some(compression_level);
         self
     }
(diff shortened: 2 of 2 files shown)


zaeleus:
Thanks for catching this! I also changed the other related builders to be owned/consuming builders in 0479ce690aec21523abcb81fb1f5bf8d0d86d981.

jamestwebber:
Happy to do it 😄 

Do you think you could share the code for the benchmarks you posted in #17? My own code is not running as fast as I expected and I am wondering if I am misusing async in some way (like blocking too much and losing parallelism).

The use-case I have here is reading a BAM, pulling out a subset that contain a certain tag value, adding a new tag and writing to a new BAM file. I have to round-trip through a SAM record, but I would think that should be very quick relative to the decompression/recompression steps.

zaeleus:
The programs used in that benchmark don't diverge much from [the examples], but note that those benchmarks are testing raw throughput. The input is read from disk, and the output goes to a sink that discards the data.

> I would think that should be very quick relative to the decompression/recompression steps.

_Relative_ is the key here. Compression and decompression are no longer on the main thread and have no weight in the comparison. In this case, I suspect your program isn't heavily I/O bound anymore; you'll have to profile your own program to find the bottleneck. :)

[the examples]: https://github.com/zaeleus/noodles/tree/cd826821206d6a504b4404da2cf1fa746ba4e1df/noodles-bgzf/examples

jamestwebber:
Hm it seems like the round-trip through the SAM record is actually quite a bottleneck--if I just read and write BAM records it is about 10x faster.

Likely the issue is that I'm doing the conversion in the main loop, and I need to learn how to write async code so that I can do it in another thread. Normally I would reach for `rayon` but I probably don't want to mix the two unless I need to. Time to read up on `tokio`

PR #37, opened 2021-09-28
Expose types in public interfaces

MaltheSR:
I was playing around with some extensions to the `vcf`/`bcf` modules, and found that certain things are made difficult by the fact that some returned structs (`Records` and `Query` in `noodles-{vcf,bcf}`) are not exported, and hence cannot be referenced from outside `noodles`. I see that similar structs are exported in other parts of `noodles`, so I assume this is unintentional and made this small PR to fix.

For completeness sake, I started going through some of the other sub-crates to see if I could find other instances of the same pattern. I can find it in a few other places (e.g. `CompressionHeader`, `Container` in `noodles-cram`), but there I'm less sure if for some reason this is intentional, and, if not, what export structure you might prefer. Happy to help if you want to outsource something.

Thanks as always for your work on this!

### DIFF
--- a/noodles-bcf/src/lib.rs
+++ b/noodles-bcf/src/lib.rs
@@ -10,7 +10,11 @@ mod reader;
 mod record;
 mod writer;
 
-pub use self::{reader::Reader, record::Record, writer::Writer};
+pub use self::{
+    reader::{Query, Reader, Records},
+    record::Record,
+    writer::Writer,
+};
 
 #[cfg(feature = "async")]
 pub use self::r#async::Reader as AsyncReader;
--- a/noodles-bcf/src/lib.rs
+++ b/noodles-bcf/src/lib.rs
@@ -6,15 +6,11 @@
 mod r#async;
 
 pub mod header;
-mod reader;
+pub mod reader;
 mod record;
 mod writer;
 
-pub use self::{
-    reader::{Query, Reader, Records},
-    record::Record,
-    writer::Writer,
-};
+pub use self::{reader::Reader, record::Record, writer::Writer};
 
 #[cfg(feature = "async")]
 pub use self::r#async::Reader as AsyncReader;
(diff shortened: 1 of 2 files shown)


zaeleus:
Thanks for bringing this up! The reader iterators are actually public but are made inaccessible due to the parent `reader` modules being private. Can you rebase and simply make `noodles_{vcf,bcf}::reader` public (i.e., `pub mod reader;` in `lib.rs`)? I prefer the types to remain namespaced.

> For completeness sake, I started going through some of the other sub-crates to see if I could find other instances of the same pattern. I can find it in a few other places (e.g. CompressionHeader, Container in noodles-cram), but there I'm less sure if for some reason this is intentional, and, if not, what export structure you might prefer.

`noodles_cram::data_container::CompressionHeader` should be public. I'll work on this one since it requires extra documentation in the submodules.

`noodles_cram::Container` is intentionally private. I don't see it leaking (in visibility), but if it is, that's a bug!

Please submit an issue for discussion when there's something that should be considered public. Adding to the public API is easy, but doing the reversal or making changes requires a lot more effort for everyone involved.

MaltheSR:
Sure, I rebased and force pushed with the changes. Apologies for not opening an issue first.

`noodles_cram::Container` is publicly visible in [`impl TryFrom<Container> for DataContainer`](https://docs.rs/noodles/0.9.0/noodles/cram/struct.DataContainer.html#impl-TryFrom%3CContainer%3E). I guess since it's never returned anywhere, though, your downstream consumers may never need to name it, so perhaps it's less of a problem. I just noticed it in the docs.

zaeleus:
Thanks!

> Apologies for not opening an issue first.

No need to apologize; I didn't mean not to submit PRs. I'm happy to receive any communication via issue/PR/email.

> `noodles_cram::Container` is publicly visible in [`impl TryFrom<Container> for DataContainer`](https://docs.rs/noodles/0.9.0/noodles/cram/struct.DataContainer.html#impl-TryFrom%3CContainer%3E).

Ah, yes, because it's a trait implementation. This is now removed from the public API in ee9e2ab529d94c706b7e9f798e34a1f6d137aa4e.

PR #39, opened 2021-09-30
sam/record: Add mutable getter for flags.

### LINKED ISSUE #38: No way to mutate the flags in a SAM record
The `Recod::flags` function returns `Flags` which is `Copy` so there is, as far as I can see, no way to mutate the flags.

```rust
pub fn flags_mut(&mut self) -> &mut Flags {
    &mut self.flags
}
```
is probably the way to go considering similar functions of the API. I can send a PR if you like.

veldsla:
Fixes #38


zaeleus:
Thanks!

PR #43, opened 2021-10-08
Access genotype alleles (Fixes #42)

### LINKED ISSUE #42: Using Genotype
I can parse a VCF record to get the `FORMAT/GT` fields as a `noodles_vcf::record::genotype::field::value::genotype::Genotype`, which I see is a newtype around `Vec<noodles_vcf::record::genotype::field::value::genotype::allele::Allele>`. However, I cannot find a way to work with these alleles in any way. From other parts of noodles I might have expected that `Genotype` implemented e.g. `Deref<Target = Vec<Allele>>` or provided getters/setters/iterators, but I cannot find any way to access the inner information. Am I missing something, or is such access lacking?

A rather length MWE:

```rust
use std::io;

use noodles_vcf as vcf;

fn main() -> io::Result<()> {
    let data = br#"##fileformat=VCFv4.1
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	NA00001	NA00002	NA00003
1	1	.	A	.	999	PASS	.	GT	0/0	./0	1/1"#;

    let mut reader = vcf::Reader::new(&data[..]);

    let _header = reader.read_header()?;

    let mut raw_record = String::new();
    reader.read_record(&mut raw_record)?;

    let record: vcf::Record = raw_record.parse().expect("cannot parse record");

    let genotypes: Vec<vcf::record::genotype::field::value::genotype::Genotype> = record
        .genotypes()
        .iter()
        .map(|gt| {
            gt.get(&vcf::record::genotype::field::key::Key::Genotype)
                .and_then(|field| field.value())
                .map(|value| match value {
                    vcf::record::genotype::field::Value::String(s) => s
                        .parse::<vcf::record::genotype::field::value::genotype::Genotype>()
                        .expect("cannot parse genotype"),
                    _ => panic!("no genotype for sample"),
                })
                .expect("cannot access genotype")
        })
        .collect();

    for genotype in genotypes {
        println!("{:?}", genotype);

        // Do something with alleles in genotype?
    }

    Ok(())
}
```

On a somewhat related note, would it be in scope for noodles to provide an iterator over `noodles_vcf::record::genotype::field::value::genotype::Genotype` from a `noodles_vcf::record::Record` in addition to the iterator over `noodles_vcf::record::genotype::Genotype` that is returned by `Record::genotypes`? For my own part at least, the `GT` field itself is by far the most commonly accessed `FORMAT` field, so this might be a nice ergonomic improvement as compared to the above.

Again, I'm happy to contribute PRs for either of those suggestions if you'd like.

MaltheSR:
Here were the basics discussed in #42. I was thinking the following might also be desirable:

1. Implement mutable getters for `position` and `phasing` for `Allele`.
2. Implement a public constructor for `Genotype`, as `pub fn new(alleles: Vec<Allele>) -> Self` and/or `From<Vec<Allele>> for Genotype`.
3. Implement `DerefMut<Target = [Allele]> for Genotype`.

However, I suppose these depend on how you envision `Genotype` being used in practice, so I thought you'd perhaps like to give some input on those before I forged ahead.

### DIFF
--- a/noodles-vcf/src/record/genotype/field/value/genotype.rs
+++ b/noodles-vcf/src/record/genotype/field/value/genotype.rs
@@ -4,7 +4,7 @@ pub mod allele;
 
 pub use self::allele::Allele;
 
-use std::{error, fmt, ops::Deref, str::FromStr};
+use std::{convert::TryFrom, error, fmt, ops::Deref, str::FromStr};
 
 /// A VCF record genotype value.
 #[derive(Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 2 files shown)


zaeleus:
1 and 3 are fine, if you'd like to do them.

For 2, I'd recommend just the fallible conversion `TryFrom<Vec<Allele>> for Genotype`, since we know the first allele cannot have phasing information. See, e.g., [`impl TryFrom<Vec<Field>> for vcf::record::Genotype`](https://github.com/zaeleus/noodles/blob/a9d3a6767ebc792aef78be91f2cdb30ec1700097/noodles-vcf/src/record/genotype.rs#L106-L154) as a reference of how to structure the error type and implementation.

MaltheSR:
Good point, thank you. I also figure a genotype with no alleles should not be representable, so I ruled that one out too.

The three latest commits should cover points 1. through 3. above.

PR #60, opened 2021-11-24
bam/writer, sam/writer: Validate cigar/sequence/basequalities

### LINKED ISSUE #59: Bam record should validate sequence and quality length before writing
In the encoded bam there is one field that encodes the length for both the sequence and the base qualities. When writing a bam::Record this is not validated resulting in possibly corrupt BAM files.

According to the [bam spec](https://samtools.github.io/hts-specs/SAMv1.pdf#subsubsection.4.2.3) when base qualities are omitted 0xFF should be written instead.

This can be done in the writer code, but it's probably cleaner to do a `validate` function on the record? Should I do a PR?

veldsla:
To avoid writing corrupt or incompatible SAM/BAM files an extra
validation step is applied before writing. This makes sure:
 - The number of base qualities is equal to the sequence length or empty
 - The length of the cigar operations matches the sequence length

Closes #59

### DIFF
--- a/noodles-bam/src/async/writer.rs
+++ b/noodles-bam/src/async/writer.rs
@@ -7,7 +7,7 @@ pub use self::builder::Builder;
 use std::ffi::CString;
 
 use noodles_bgzf as bgzf;
-use noodles_sam as sam;
+use noodles_sam::{self as sam, validate};
 use tokio::io::{self, AsyncWrite, AsyncWriteExt};
 
 use crate::Record;
@@ -167,6 +167,7 @@ where
         reference_sequences: &sam::header::ReferenceSequences,
         record: &sam::Record,
     ) -> io::Result<()> {
+        validate(record)?;
         sam_record::write_sam_record(&mut self.inner, reference_sequences, record).await
     }
 }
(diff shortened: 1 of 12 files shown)


veldsla:
This probably needs renaming or moving of stuff. Let me know what you think!

zaeleus:
Sorry, this is a bit too out of scope of what was initially suggested in #59. I expected the same logic used in the (BAM) SAM record writer to be ported to the BAM record writer. Can you scale it back to just that?

veldsla:
I agree the file corruption issue is a different problem from writing sam/bam files that do not conform to the specification. Unfortunately the record writing code is duplicated 6 times. This way the validation is only written twice.

Fixing only the length/corruption issue does not magically make the incompatibility issue go away.

veldsla:
Apparently using this branch on real world bam files triggers errors, but not in the test suite. Will investigate, converted PR to WIP.

zaeleus:
Thanks for looking at this in the past. I'm closing this since the alignment parsers/writers have now diverged greatly and makes the same checks.

PR #61, opened 2021-11-25
writer/record, async/writer/record: Avoid file corruption

### LINKED ISSUE #59: Bam record should validate sequence and quality length before writing

veldsla:
To avoid file corruption verify that the length of sequence matches the
length of the quality scores.

This only fixes the potential file corruption issue. The SAM specification issues remaining are:
 - sam: The sam writer can write sam lines with quality score length differing from seq length (but noodles-sam reader cannot parse it)
 - sam/bam: The cigar operations length must match the sequence length. This is not validated and when written incorrectly results in incompatible sam/bam files

These issues are addressed in #60 and that pull request is mutually exclusive with this one.

Closes #59

### DIFF
--- a/noodles-bam/src/async/writer/record.rs
+++ b/noodles-bam/src/async/writer/record.rs
@@ -1,6 +1,6 @@
 use tokio::io::{self, AsyncWrite, AsyncWriteExt};
 
-use crate::Record;
+use crate::{Record, NULL_QUALITY_SCORE};
 
 pub(super) async fn write_record<W>(writer: &mut W, record: &Record) -> io::Result<()>
 where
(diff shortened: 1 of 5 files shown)


zaeleus:
Thanks!

PR #63, opened 2021-12-17
noodles-sam: fix reader to parse some edge cases

veldsla:
I was playing with the tests for SAM defined in the [hts-specs repo](https://github.com/samtools/hts-specs/tree/master/test/sam).

I used a simple [bash script](https://gist.github.com/veldsla/260e8f5f7e09cb82e73fba6348ed9158) and the sam_view example to read the files.

There are two sets, passed and failed.
The fail tests are a bit odd. Most also don't fail a `samtools view` so it's difficult to decide.

All the pass test parse with `samtools view`, but a couple don't with `sam_view`

This PR fixes 2 of those edge cases:
 - The hex value in a field maybe in lowercase
 - A cigar line maybe missing (*) but can the read can still be mapped and have a sequence (this used to trigger a `SequenceLengthMismatch`)

Now 2 test still fail to parse, but it seems this is an issue with the test (duplicate tags).

The bash script also compares the output of the view commands and there are some differences in the handling between noodles and samtools:
 - Samtools replaces invalid bases in the sequence with N
 - Samtools adjust the flags when it decides the read cannot be mapped based on other fields

Not sure what to to with this.

-edit And of course I didn't run the tests...shame on me. Removed the expected error on the lowercase hex `cafe`. Was also a bit strange because it's even in the examples somewhere:
`assert_eq!(Value::Hex(String::from("cafe")).as_hex(), Some("cafe"));`

### DIFF
--- a/noodles-sam/src/record/builder.rs
+++ b/noodles-sam/src/record/builder.rs
@@ -325,7 +325,7 @@ impl Builder {
     pub fn build(self) -> Result<Record, BuildError> {
         // § 1.4 The alignment section: mandatory fields (2021-06-03): "If not a '*', the length of
         // the sequence must equal the sum of lengths of `M/I/S/=/X` operations in `CIGAR`."
-        if !self.flags.is_unmapped() && !self.sequence.is_empty() {
+        if !self.flags.is_unmapped() && !self.sequence.is_empty() && !self.cigar.is_empty() {
             let sequence_len = self.sequence.len() as u32;
             let cigar_read_len = self.cigar.read_len();
 
--- a/noodles-sam/src/record/data/field/value.rs
+++ b/noodles-sam/src/record/data/field/value.rs
@@ -791,7 +791,6 @@ mod tests {
         assert_eq!("Z:🍜".parse::<Value>(), Err(ParseError::InvalidStringValue));
 
         assert_eq!("H:CAFE".parse(), Ok(Value::Hex(String::from("CAFE"))));
-        assert_eq!("H:cafe".parse::<Value>(), Err(ParseError::InvalidHexValue));
         assert_eq!("H:CAFE0".parse::<Value>(), Err(ParseError::InvalidHexValue));
         assert_eq!("H:NDLS".parse::<Value>(), Err(ParseError::InvalidHexValue));
 
--- a/noodles-sam/src/record/data/field/value.rs
+++ b/noodles-sam/src/record/data/field/value.rs
@@ -642,7 +642,7 @@ fn parse_string(s: &str) -> Result<String, ParseError> {
 
 // § 1.5 The alignment section: optional fields (2021-01-07)
 fn is_valid_hex_char(c: char) -> bool {
-    matches!(c, '0'..='9' | 'A'..='F')
+    matches!(c, '0'..='9' | 'A'..='F' | 'a'..='f')
 }
 
 fn parse_hex(s: &str) -> Result<String, ParseError> {


zaeleus:
Thanks for testing!

> The hex value in a field maybe in lowercase

Hex field values must be uppercase. This is defined by [§ 1.5 in the spec]: `([0-9A-F][0-9A-F])*`.

> A cigar line maybe missing (*) but can the read can still be mapped and have a sequence

The behavior for this seems to be undefined. It contradicts the htslib warning `[W::sam_parse1] mapped query must have a CIGAR; treated as unmapped`, which implies mapped queries are required to have CIGAR operations.

> Samtools replaces invalid bases in the sequence with N

This behavior is undefined. Invalid bases are accepted in [§ 1.4]: `\*|[A-Za-z=.]+`.

> Samtools adjust the flags when it decides the read cannot be mapped based on other fields

This behavior contradicts the spec: "Bit 0x4 is the only reliable place to tell whether the read is unmapped." If the unmapped flag is not set, using other fields to infer it cannot make the same guarantee.

[§ 1.5 in the spec]: https://github.com/samtools/hts-specs/blob/6835cd374bc2b6a0c6ac88aa2faff4880a5edbc0/SAMv1.tex#L661
[§ 1.4]: https://github.com/samtools/hts-specs/blob/6835cd374bc2b6a0c6ac88aa2faff4880a5edbc0/SAMv1.tex#L422

veldsla:
> > The hex value in a field maybe in lowercase
> 
> Hex field values must be uppercase. This is defined by [§ 1.5 in the spec](https://github.com/samtools/hts-specs/blob/6835cd374bc2b6a0c6ac88aa2faff4880a5edbc0/SAMv1.tex#L661): `([0-9A-F][0-9A-F])*`.

Yes, it is indeed, another issue with the tests then. If it were my choice I'd probably change the spec. Nothing wrong with lowercase hex right?

> > A cigar line maybe missing (*) but can the read can still be mapped and have a sequence
> 
> The behavior for this seems to be undefined. It contradicts the htslib warning `[W::sam_parse1] mapped query must have a CIGAR; treated as unmapped`, which implies mapped queries are required to have CIGAR operations.

You are right. When viewing the file the flag is adjusted and the warning is printed. so this is related to the last point

> > Samtools replaces invalid bases in the sequence with N
> 
> This behavior is undefined. Invalid bases are accepted in [§ 1.4](https://github.com/samtools/hts-specs/blob/6835cd374bc2b6a0c6ac88aa2faff4880a5edbc0/SAMv1.tex#L422): `\*|[A-Za-z=.]+`.

Agreed, this might be because the internal representation is bam which forces the conversion.
> 
> > Samtools adjust the flags when it decides the read cannot be mapped based on other fields
> 
> This behavior contradicts the spec: "Bit 0x4 is the only reliable place to tell whether the read is unmapped." If the unmapped flag is not set, using other fields to infer it cannot make the same guarantee.

This is from the [recommendations]( https://samtools.github.io/hts-specs/SAMv1.pdf#section.2), but indeed not part of the spec:
> If POS plus the sum of lengths of M/=/X/D/N operations in CIGAR exceeds the length specified in
the LN field of the @SQ header line (if exists) with an SN equal to RNAME, the alignment should
be unmapped, unless the reference sequence is circular

It seems the issues are indeed more with the spec/test-data and can better be raised there. Will close this for now, but keep the branch in case the specs change. Too bad there isn't a ready test set to validate an implementation

PR #66, opened 2022-01-05
Add mutable access to VCF header

### LINKED ISSUE #65: Mutable access to VCF header
As far as I can tell, it's currently impossible to modify an existing `vcf::Header`, apart from adding unstructured records with the `insert` method. This is inconvenient e.g. for the case when modifying an existing VCF by adding a few INFO or FORMAT fields to the header and records, rather than building a header from scratch.

I see that the `sam::Header` already has `_mut` methods for cases like this. Would it make sense to add something similar for the `vcf::Header` methods that return `IndexMap`s? For the non-`IndexMap` fields (`assembly`, `file_format`, `pedigree_db`), there could be similar `_mut` methods to return mutable reference or perhaps simply `set_` setters? Finally, it might sometimes be ergonomic to create a `vcf::header::Builder` from an existing header - perhaps `impl From<Header> for Builder` could be nice?

Happy to make the PR if you'd like.

malthesr:
Adds mutable getters for all structured fields of a VCF header. I also fixed a couple of typos in related, existing doc comments to make things match up.

Two points:
1. I didn't add doc tests for the new methods. I figured that they might make the doc pages unnecessarily long while being more or less self-explanatory together with the existing doc tests, but perhaps they'd still be good to have. Happy to back over and add these if you prefer? 
2. The header is still lacking mutable access to the unstructured fields. Perhaps it would make sense to add either `get_mut`/`delete` to go with `get`/`insert`, and/or simply add `records`/`records_mut` for full access? 

Closes #65

### DIFF
--- a/noodles-vcf/CHANGELOG.md
+++ b/noodles-vcf/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+  * vcf/header: Add mutable getters for `vcf::Header` fields.
+
 ## 0.11.1 - 2021-12-09
 
 ### Fixed
--- a/noodles-vcf/src/header.rs
+++ b/noodles-vcf/src/header.rs
@@ -100,6 +100,13 @@ impl Header {
         self.file_format
     }
 
+    /// Returns a mutable reference to the file format (`fileformat`) of the VCF.
+    ///
+    /// `fileformat` is a required meta record and is guaranteed to be set.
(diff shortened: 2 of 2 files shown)


zaeleus:
Thanks!

1. There is no strict requirement, but always prefer to write examples, even it is seems duplicated or obvious. It's better to avoid the user from hunting for examples.
2. Agree and tracking in https://github.com/zaeleus/noodles/issues/68. Feel free to add it if it's unassigned!

PR #67, opened 2022-01-05
Add mutable access to VCF record

malthesr:
I came across a situation where I needed to mutate the genotypes of existing VCF records. As in #65, I figured it'd be good to have `genotypes_mut`. In the process, it seemed natural to also implement `reference_bases_mut` and `alternate_bases_mut` as well as `DerefMut` for `ReferenceBases` and `AlternateBases` to also have mutable access to these parts of the record.

EDIT: Realised mutating the genotypes of an existing record required a bit more than I though. Implemented `DerefMut` for `Genotype` and `Keys` and a mutable getter for `Keys` also. As always, feel free to disregard or request changes as you see fit.

### DIFF
--- a/noodles-vcf/CHANGELOG.md
+++ b/noodles-vcf/CHANGELOG.md
@@ -12,9 +12,15 @@
   * vcf/record: Add mutable getters for genotypes, reference bases, and
     alternate bases.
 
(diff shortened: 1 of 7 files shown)


zaeleus:
Thanks!

Great work keeping the changes small and incremental. It really helps when reviewing.

PR #69, opened 2022-01-05
vcf/header: Add getter and mutable getter for unstructured records

### LINKED ISSUE #68: vcf/header: Add generic records accessors
> 2\. The header is still lacking mutable access to the unstructured fields. Perhaps it would make sense to add either `get_mut`/`delete` to go with `get`/`insert`, and/or simply add `records`/`records_mut` for full access?

https://github.com/zaeleus/noodles/pull/66

---

Add

  * `vcf::Header::records -> &IndexMap<String, Vec<Record>>` and
  * `vcf::Header::records_mut -> &mut IndexMap<String, Vec<Record>>`

to access the header `map` field. This is a map of generic records.

malthesr:
Closes #68

### DIFF
--- a/noodles-vcf/CHANGELOG.md
+++ b/noodles-vcf/CHANGELOG.md
@@ -6,6 +6,9 @@
 
   * vcf/header: Add mutable getters for `vcf::Header` fields.
 
+  * vcf/header: Add immutable and mutable getters for unstructured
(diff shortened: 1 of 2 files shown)


zaeleus:
Nice, thanks!

PR #70, opened 2022-01-13
vcf/record: Derive PartialOrd/Ord for newtypes

malthesr:
Just a little ergonomic improvement, adding derives

- `PartialOrd`/`Ord` for `vcf::record::Position`
- `PartialOrd` for `vcf::record::QualityScore`

to avoid having to go in and out of the wrapped type for comparisons.


zaeleus:
Thanks!

PR #73, opened 2022-02-04
A fastx reader

natir:
In this pull request I add a fastx (fasta|fastq) parser. 

To perform this I create a struct `Record` to store information of fastq or fasta record, this record perform parsing from a `std::io::BufRead` and writing in `std::io::Write`. I use this struct to create a `Reader`, `Writer` and a `Records` iterator.

To make difference between Fasta and Fastq I just use the prefix of comment line (`>`, `@`).

I compare my parser with noodles_fasta and noodles_fastq, performance are quite similar.  

I make this pull request, before write documentation, to be sure naming/styling/api and other stuff like these are compatible with noodles standard. 

Todo list:
- [x] Record
- [x] Reader
- [x] Records
- [x] Writer
- [x] Test
- [x] Benchmark
- [x] Fix some performance/style issue
- [ ] Documentation


zaeleus:
Can you explain the motivation behind this? The example (`fastq2fasta`) can be achieved using the FASTA and FASTQ crates, and I'm not familiar with inputs having mixed record formats (`noodles-fastx::reader::tests::read_record`).

natir:
The main motivation is, have tools able to read a sequence file without:
- check extension of file (extension of file lie)
- check begin of file
- care about if it's fastq or fasta 

The more important point is the last one, I write many tools don't care about quality string but it nice to let user put fastq file as input. Create a copy of sequence file without quality string, is annoying, take time and disk space. In some situation user can perform conversion and pipe result in stdin but it's not the case in many situation.

Mixed record formats it's just a nice side effect of how I manage parsing of fast·a·q in same time.

natir:
Hi @zaeleus,

In last commit, I to simplify the most complex part of code.

I understand your concern about motivation behind this type of parser, and I think in previous comment I show you some good argument but I think the best argument is I need this type parser.

zaeleus:
Thanks for the contribution, @natir.

After some consideration, I don't think this is a good fit to include in noodles. FASTX is not a format; it represents an abstraction over two existing ones. This does not warrant a new package.

There are currently no plans to add unified I/O.

You may want to move this into a separate library or use existing options, e.g., [needletail](https://github.com/onecodex/needletail), [seq_io](https://github.com/markschl/seq_io), etc.

natir:
Hi @zaeleus,

I like the idea of noodles crate, one crate/features per file format, and your point is clear and logic. 

If you plan concern unified I/O change in future, don't hesitate to use this code.

PR #82, opened 2022-04-10
add support for bed12

pollend:
here is a work in progress that add bed12 support to the parser. was curious if it would be nice to use traits to specify how the additional settings would be parsed. there are a couple variations listed out here: http://genome.ucsc.edu/FAQ/FAQformat#format1

```
Bed6<Modifier1>
Bed12<Modifier2>
etc..
```

### DIFF
--- a/noodles-bed/src/record.rs
+++ b/noodles-bed/src/record.rs
@@ -1,11 +1,12 @@
 //! BED record and fields.
 
 pub mod builder;
+pub mod color;
 pub mod name;
 pub mod score;
 pub mod strand;
 
-pub use self::{builder::Builder, name::Name, score::Score, strand::Strand};
+pub use self::{builder::Builder, color::Color, name::Name, score::Score, strand::Strand};
 
 use std::{
     error,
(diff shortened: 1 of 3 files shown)


noodles-bed/src/record.rs
```diff
@@ -29,6 +32,11 @@ struct StandardFields {
     name: Option<Name>,
     score: Option<Score>,
     strand: Option<Strand>,
+    thick_start: Option<Position>,
+    thick_end: Option<Position>,
+    item_rgb: Option<Color>,
+    block_sizes: Option<Vec<usize>>,
```
pollend, line 38: would this preferable to make this a type Size.

zaeleus:
Thanks for your interest, @pollend!

I actually have a significant portion of BED{7,8,9,12} completed in a local branch. Do you use the display and/or block fields in BED that require them to be added?

pollend:
> Thanks for your interest, @pollend!
> 
> I actually have a significant portion of BED{7,8,9,12} completed in a local branch. Do you use the display and/or block fields in BED that require them to be added?

by display and block fields are you referring to 9 color (10 11 and 12) for the list of blocks. If so yes, I have verified those in a test case.

zaeleus:
Sorry, I meant what is your motivation behind the PR? Are you building an application that uses all of the fields?

pollend:
> Sorry, I meant what is your motivation behind the PR? Are you building an application that uses all of the fields?

nothing in particular, just wanted to get into a bit of bioinformatics and I like rust.

zaeleus:
I merged in [prior work](https://github.com/zaeleus/noodles/compare/ac2916c6c4...aa888dbcd3) for BED{7,8,9,12}+ support, but thanks for the interest!

PR #81, opened 2022-04-06
bugfix: resolve conversion from char to us for base

pollend:
I was trying to build from nightly and From<char> is not implement for u8. this is a primative type so you can simply do a straight conversion to u8.

```
error[E0277]: the trait bound `u8: From<char>` is not satisfied
  --> noodles-sam/src/record/sequence/base.rs:89:9
   |
89 |         u8::try_from(c)
   |         ^^^^^^^^^^^^^^^ the trait `From<char>` is not implemented for `u8`
   |
   = help: the following implementations were found:
             <u8 as From<CompressionLevel>>
             <u8 as From<NonZeroU8>>
             <u8 as From<base::Base>>
             <u8 as From<bool>>
           and 4 others
   = note: required because of the requirements on the impl of `Into<u8>` for `char`
   = note: required because of the requirements on the impl of `TryFrom<char>` for `u8`
```

PR #102, opened 2022-07-19
Update Cargo.toml for noodles-sam

nh13:
I think this is needed for [`TryFrom<char> for u8`](https://doc.rust-lang.org/std/primitive.u8.html#impl-TryFrom%3Cchar%3E) in [base.rs#L89](https://github.com/zaeleus/noodles/blob/75da943f5fa8a098866edf7d4a49572039e11cad/noodles-sam/src/record/sequence/base.rs#L89).  See my comments on #81.


zaeleus:
Ah, thanks for tracking #81 down. I was never able to reproduce it and wasn't aware `TryFrom<char> for u8` was only added in Rust 1.59.0.

The alternative, more conservative approach would be to replace https://github.com/zaeleus/noodles/blob/75da943f5fa8a098866edf7d4a49572039e11cad/noodles-sam/src/record/sequence/base.rs#L89 with `u8::try_from(u32::from(c))` instead. This would not require a MSRV change. Otherwise, we'll have to bubble this change to dependent crates: noodles-bam, noodles-cram, noodles-util, and noodles.

Do you have a preference?

nh13:
The latter is more permissive, so I’d pick the more conservative approach.

zaeleus:
Sounds good. Feel free to close this and open a new PR with that change.

jch-13:
A quick run of [`cargo-msrv`](https://crates.io/crates/cargo-msrv) returned that the current MSRV is `v1.57.0` due to
```
error[E0658]: panicking in constants is unstable                                                                                                                                           │
│   --> noodles-core/src/position.rs:21:17                                                                                                                                                   │
│    |                                                                                                                                                                                       │
│ 21 |         None => unreachable!(),                                                                                                                                                       │
│    |                 ^^^^^^^^^^^^^^
```
I wonder if it's fine to set the MSRV to `v1.59.0` if the version has to be raised anyway.

nh13:
I’d just go with this PR then. It’s not a big burden to update.

zaeleus:
> A quick run of [`cargo-msrv`](https://crates.io/crates/cargo-msrv) returned that the current MSRV is `v1.57.0`

Well then. Thanks for checking!

I have no problem raising the MSRV to 1.59.0. I'll merge this PR and update the rest of the crates afterward.

PR #105, opened 2022-07-27
Add GZI

### LINKED ISSUE #103: BGZF Index
Is there any support for reading BGZF index files, that is, [.gzi files](http://www.htslib.org/doc/bgzip.html#GZI_FORMAT)? If not, is there any interest in this feature? I'd be happy to help/contribute with writing something up.

For [htsget-rs](https://github.com/umccr/htsget-rs) we are thinking of using a bgzf index to supplement other indices. This could avoid some reads/seeks to the underlying file (the file the indices are for) for certain htsget queries.

mmalenic:
Closes #103

The following PR implements a GZ Index and Reader for BGZF files.
* Adds a GZI module containing a GZ Index type for pairs of offsets.
* Implements a GZ Index Reader.

### DIFF
--- a/noodles-bgzf/src/gzi.rs
+++ b/noodles-bgzf/src/gzi.rs
@@ -1,5 +1,5 @@
 //! A module for the [GZI Index]. A GZI index contains pairs of compressed and uncompressed offsets
-//! in a BGZF file, where values are stored as little-endian 64-bit unsigned integers.
+//! in a BGZF file. Values in the index are stored as little-endian 64-bit unsigned integers.
 //!
 //! [GZI Index]: http://www.htslib.org/doc/bgzip.html#GZI_FORMAT
 
--- a/noodles-bgzf/src/gzi.rs
+++ b/noodles-bgzf/src/gzi.rs
@@ -2,5 +2,8 @@
 
 mod reader;
 
-/// A GZI Index, representing pairs of compressed and uncompressed offsets in a BGZF file.
-pub type Index = Vec<(u64, u64)>;
+/// A GZI Index containing a number of entries, representing pairs of compressed and uncompressed offsets in a BGZF file.
+pub struct Index {
+  number_entries: u64,
+  offsets: Vec<(u64, u64)>
+}
--- a/noodles-bgzf/src/gzi.rs
+++ b/noodles-bgzf/src/gzi.rs
@@ -1,0 +1,6 @@
+//! GZI index.
+
+mod reader;
+
+/// A GZI Index, representing pairs of compressed and uncompressed offsets in a BGZF file.
+pub type Index = Vec<(u64, u64)>;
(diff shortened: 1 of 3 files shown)


noodles-bgzf/src/gzi.rs
```diff
@@ -0,0 +1,79 @@
...
+//! [GZI Index]: http://www.htslib.org/doc/bgzip.html#GZI_FORMAT
+
+pub use self::reader::Reader;
+
+mod reader;
+
+/// A GZI Index contains a number of entries, representing pairs of compressed and uncompressed
+/// offsets in a BGZF file.
+#[derive(Debug, PartialEq)]
+pub struct Index {
```
zaeleus, line 13: I suggested creating a type alias for `Index` as `Vec<(u64, u64)>` in https://github.com/zaeleus/noodles/issues/103#issuecomment-1190472420 since a wrapper like this doesn't improve the model, e.g., `number_entries` is redundant given it's the same as `offsets.len()`.

You could create a newtype, but I'm not sure if it'd be useful.

noodles-bgzf/src/gzi/reader.rs
```diff
@@ -0,0 +1,158 @@
...
+    reader.read_exact(&mut buf)?;
+    Ok(u64::from_le_bytes(buf))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_read_index() -> Result<(), Box<dyn std::error::Error>> {
```
zaeleus, line 81:
```suggestion
    fn test_read_index() -> io::Result<()> {
```

Et al.

noodles-bgzf/src/gzi/reader.rs
```diff
@@ -0,0 +1,158 @@
...
+            0x02, 0x5d, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // compressed_offset = 23810
+            0x01, 0x52, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, // uncompressed_offset = 86529
+        ];
+
+        let mut reader = Reader::new(&data[..]);
+
+        let actual = reader.read_index().map_err(|err| err.kind());
+        let expected = Err(io::ErrorKind::UnexpectedEof);
+
+        assert_eq!(actual, expected);
```
zaeleus, line 135:
```suggestion
        assert!(matches!(
            reader.read_index(),
            Err(e) if e.kind() == io::ErrorKind::UnexpectedEof
        ));
```

Et al.

zaeleus (changes requested):
Doesn't GZI mean GZ index, making GZI index redundant?

noodles-bgzf/src/gzi.rs
mmalenic: Yeah, you're right. I think the only benefit might be wrapping `offsets.len()` in a function which represents the definition of GZI, e.g:
```rust
  pub fn number_entries(&self) -> u64 {
      self.0.len() as u64
  }
```
Although, that might be complicating things more than necessary.

mmalenic:
Thanks @zaeleus, I've made those changes.

mmalenic:
I'd be interested to implement the `async` version of this too. Would implementing an `async` version belong in this PR, or another?

zaeleus (approved):

zaeleus:
Thanks!

PR #106, opened 2022-07-29
Add async GZI

mmalenic:
Similar to #103 and #105, the following PR adds the async version of GZI.
* Adds async module in gzi module.
* Implements async gzip reader.


zaeleus (approved):

zaeleus:
Thanks again!

Issue #108, opened 2022-08-22
Region query on the bgzipped indexed fasta?


lu6huimin:
Problem: for noodles-fasta 0.13.0, query on bgzipped indexed fasta has not yet supported, right?

Possible solution: implement `std::io::Seek` for `noodles_bgzf::Reader`.

However, it will conflict with the original `fn seek(pos: VirtualPosition)`. Is that OK to rename all `seek` of virtual position to `seek_virtual_position`, or anything else you like? I can make a draft implementation, if a pull request is welcome.

lu6huimin:
@zaeleus Would you like to comment on this? For large genomes, fasta files are bgzipped in most cases.

zaeleus:
noodles-fasta can now seek/query bgzipped FASTA files. See [`fasta::IndexedReader`](https://docs.rs/noodles/0.29.0/noodles/fasta/struct.IndexedReader.html) and the [`fasta_query`](https://github.com/zaeleus/noodles/blob/e8c5fe9b1c9d4cd4f8801936ea6692397c8c589c/noodles-fasta/examples/fasta_query.rs) example.

lu6huimin:
Thanks!

PR #109, opened 2022-08-22
Implement region query for the reader of bgzipped indexed fasta

lu6huimin:
See #108, the key is to `impl Seek for bgzf::Reader`. My draft implementation has three steps:

1. Rename all `fn seek(&mut self, pos: VirtualPosition)` to `fn seek_virtual_position(&mut self, pos: VirtualPosition)`.
2. `impl Seek for bgzf::Reader`.
* The reader is only seekable when it has the gzi to refer. So I added two fields `gzi: Option<gzi::Index>` and `uncompressed_position: Option<u64>`, to store the index and uncompressed offset, respectively. Plus a help function `bgzf::Reader::with_gzi` to initialize a seekable reader.

* The `uncompressed_position` is relative to the start of the uncompressed **file**, so it should move forward after calling `fn consume`. This field is different from the `upos` in `VirtualPosition` which is relative to the uncompressed data in the **block**. I implemented conversion from `uncompressed_position` to `VirtualPosition` by referring the [answer](https://github.com/samtools/htslib/issues/473#issuecomment-281093049), to make use of the `fn seek_virtual_position` in `fn seek`.
3. Add an example of querying bgzipped indexed fasta file. 

Feel free to modify the added code, or discard the changes.

BTW, thanks for building this fantastic crate.

### DIFF
--- a/noodles-bam/src/async/reader.rs
+++ b/noodles-bam/src/async/reader.rs
@@ -387,12 +387,12 @@ where
     /// let mut reader = bam::AsyncReader::new(Cursor::new(data));
     ///
     /// let virtual_position = bgzf::VirtualPosition::default();
-    /// reader.seek(virtual_position).await?;
+    /// reader.seek_virtual_position(virtual_position).await?;
     /// # Ok(())
     /// # }
     /// ```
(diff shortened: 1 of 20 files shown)


lu6huimin:
I recommitted to avoid `unwrap` in `test_seek`, also changed "Pull request" to "Draft pull request".

New request is here #110.

PR #110, opened 2022-08-23
Implement region query for the reader of bgzipped indexed fasta

lu6huimin:
See #108, the key is to `impl Seek for bgzf::Reader`. My draft implementation has three steps:

1. Rename all `fn seek(&mut self, pos: VirtualPosition)` to `fn seek_virtual_position(&mut self, pos: VirtualPosition)`.
2. `impl Seek for bgzf::Reader`.
* The reader is only seekable when it has the gzi to refer. So I added two fields `gzi: Option<gzi::Index>` and `uncompressed_position: Option<u64>`, to store the index and uncompressed offset, respectively. Plus a help function `bgzf::Reader::with_gzi` to initialize a seekable reader.

* The `uncompressed_position` is relative to the start of the uncompressed **file**, so it should move forward after calling `fn consume`. This field is different from the `upos` in `VirtualPosition` which is relative to the uncompressed data in the **block**. I implemented conversion from `uncompressed_position` to `VirtualPosition` by referring the [answer](https://github.com/samtools/htslib/issues/473#issuecomment-281093049), to make use of the `fn seek_virtual_position` in `fn seek`.
3. Add an example of querying bgzipped indexed fasta file. 

Feel free to modify the added code, or discard the changes.

BTW, thanks for building this fantastic crate.

### DIFF
--- a/noodles-bam/src/async/reader.rs
+++ b/noodles-bam/src/async/reader.rs
@@ -387,12 +387,12 @@ where
     /// let mut reader = bam::AsyncReader::new(Cursor::new(data));
     ///
     /// let virtual_position = bgzf::VirtualPosition::default();
-    /// reader.seek(virtual_position).await?;
+    /// reader.seek_virtual_position(virtual_position).await?;
     /// # Ok(())
     /// # }
     /// ```
-    pub async fn seek(&mut self, pos: bgzf::VirtualPosition) -> io::Result<bgzf::VirtualPosition> {
-        self.inner.seek(pos).await
+    pub async fn seek_virtual_position(&mut self, pos: bgzf::VirtualPosition) -> io::Result<bgzf::VirtualPosition> {
+        self.inner.seek_virtual_position(pos).await
     }
 
     /// Returns a stream over records that intersect the given region.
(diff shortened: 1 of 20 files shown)


zaeleus:
Sorry for the long response time, and thanks for your patience.

I don't think this is the best approach to this problem. Readers in noodles are largely agnostic to indices, and this would couple the GZ index to the BGZF reader. However, I'm still experimenting with possible solutions.

lu6huimin:
It doesn't matter, thank you for investigating on it. It's my pleasure to give any help.

I agree with you, directly implementing `Seek` for BGZF reader is not good. It's useless to seek uncompressed position for BAM which also uses BGZF reader.

Maybe re-implementing `Seek` for FASTA reader is better.

lu6huimin:
Sorry, opposite to my last reply, I find implementing `Seek` for `uncompressed position` maybe OK, although it's useless for most readers in noodles.

GZI (https://www.htslib.org/doc/bgzip.html#GZI_FORMAT) is agnostic to the uncompressed format (FASTA or anything else). It's designed for seeking uncompressed position, which looks like a basic function for BGZF reader.

Out of curiosity, I searched in `htslib` source code, finding that it also provides [`bgzf_useek`](https://github.com/samtools/htslib/blob/1.16/bgzf.c#L2499).

zaeleus:
Thanks for your interest and possible solution. A different approach was implemented by delegating to a seekable raw reader. See `{bgzf,fasta}::IndexedReader`.

PR #117, opened 2022-10-09
core/position: Add checked_sub method

claymcleod:
I need this method for some position arithmetic I'm doing. Alternatively, I can cast to a `usize`, but I'd argue this should be added since `Position` supports `add_checked`.

### DIFF
--- a/noodles-core/src/position.rs
+++ b/noodles-core/src/position.rs
@@ -59,6 +59,31 @@ impl Position {
     pub fn checked_add(self, other: usize) -> Option<Self> {
         usize::from(self).checked_add(other).and_then(Self::new)
     }
+
+    /// Subtracts an unsigned integer from a 1-based position.
+    ///
+    /// This returns `None` if the operation underflowed.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use noodles_core::Position;
+    /// let position = Position::try_from(8)?;
+    /// assert_eq!(position.checked_sub(5), Position::new(3));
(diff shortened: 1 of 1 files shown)


zaeleus:
In this case, checked subtraction is not implemented because subtraction is not arithmetically stable, i.e., it allows for a naturally occurring zero result. The checked methods are for over/underflow.

claymcleod:
You're right on this of course, but in my implementation, aren't we protected from zero `Position`s because of the call to `Self::new` (https://github.com/claymcleod/noodles/blob/position_subtract/noodles-core/src/position.rs#L40)? The behavior is the same as if I called `Position::new(0)`, which is what I would expect in this case.

It also leads to the question, if the `Position` types are not meant to support a more full set of arithmetic operations, then what is the canonical way to do them from your perspective? If it's to cast to `usize`, perhaps we should also support `From<Position>` for `NonZeroUsize`?

zaeleus:
Most arithmetic should be done with the primitive and rechecked afterward. You can do additional underflow checking if it's required by your application.

```rust
let p = Position::MIN;

let q = Position::new(usize::from(p) - other).expect("invalid position");

let n = usize::from(p).checked_sub(other).expect("attempt to subtract with overflow")
let q = Position::new(n).expect("invalid position");
```

You can also use the method in your patch as a utility function, but note that the optionals returned are not semantically the same.

claymcleod:
Okay, sounds good. I'll just use casts here then. Thanks for taking a look

PR #119, opened 2022-10-23
sam/alignment_reader: Changes `read_alignment_header` to return `String`

claymcleod:
The rationale for this change is simple: similar to the other `read_header` methods, returning a `String` allows the user to parse the `sam::Header` _after_ applying corrections to raw header text. This is desirable in cases where the header may not conform to the specification, so preprocessing of the header is required before parsing.

As this is the first change to `noodles-util` after release, I'm not sure how you want to handle changelog updates. Here, I have elected to file the changes under `noodles-sam`, since that is where the trait that was updated is located. I can also update the other changelogs if you think it's necessary (including `noodles-util`).


claymcleod:
Any thoughts on this one @zaeleus? My sense is that you are waiting to see the outcome of https://github.com/samtools/hts-specs/issues/679 before deciding if noodles should go this direction.

zaeleus:
Sorry, I wanted to experiment with a more flexible API for reading the header but don't have a complete plan yet.

In the interim for noodles-util, I'd rather `alignment::Reader::read_header` continue to return a `sam::Header`, despite this being the opposite of what `{sam,bam,cram}::Reader::read_header` do. I would propose adding `sam::AlignmentReader::read_raw_alignment_header(&mut self) -> io::Result<String>` to return the header as a string and `alignment::Reader::read_raw_header(&mut self) -> io::Result<String>` to delegate to the inner reader.

For the changelogs, yes, only update the directly affected crates, not the transitive ones.

zaeleus:
Header parsing was relaxed in noodles-sam 0.22.0 and now passes the blocker (read group platform case-sensitivity) for a majority of inputs now. The alignment format readers no longer return a raw string when reading the header. Thanks for looking into it previously.

PR #120, opened 2022-10-23
gff: Formalizes `Comment` and adds `write_comment` method to writer

claymcleod:
I suggest we formalize `Comment` as its own type and add the related `write_comment` method to the writer. The argument for the former is one of type specificity being a net positive when available (since this is not really an unconstrained `String`, better to define the parameters within which a comment can exist). The argument for the second is simply that we do not have a method for adding comments to the writer. 

I have also updated the examples:

* `gff_view.rs` has been updated to print all content of the GFF file, not just the records. I have also changed it to use the `gff::Writer` pointed at stdout.
* `gff_write.rs` has been updated to demonstrate how to write a comment.

### DIFF
--- a/noodles-gff/examples/gff_view.rs
+++ b/noodles-gff/examples/gff_view.rs
@@ -13,9 +13,13 @@ fn main() -> io::Result<()> {
 
     let mut reader = File::open(src).map(BufReader::new).map(gff::Reader::new)?;
 
-    for result in reader.records() {
-        let record = result?;
-        println!("{}", record);
+    let stdout = io::stdout();
+    let handle = stdout.lock();
+    let mut writer = gff::Writer::new(handle);
+
+    for result in reader.lines() {
+        let line = result?;
+        writer.write_line(&line)?;
     }
 
     Ok(())
--- a/noodles-gff/examples/gff_write.rs
+++ b/noodles-gff/examples/gff_write.rs
@@ -4,7 +4,7 @@
 
 use std::io;
 
-use noodles_gff as gff;
+use noodles_gff::{self as gff, Line};
 
 fn main() -> io::Result<()> {
     let stdout = io::stdout();
@@ -13,6 +13,9 @@ fn main() -> io::Result<()> {
 
     let version = gff::Directive::GffVersion(Default::default());
     writer.write_directive(&version)?;
+
+    let comment = Line::Comment(String::from("format: gff3"));
+    writer.write_line(&comment)?;
 
     let record = gff::Record::default();
     writer.write_record(&record)?;
--- a/noodles-gff/src/writer.rs
+++ b/noodles-gff/src/writer.rs
@@ -1,6 +1,6 @@
 use std::io::{self, Write};
 
-use super::{Directive, Record};
+use super::{Directive, Line, Record};
 
 /// A GFF writer.
 pub struct Writer<W> {
@@ -34,6 +34,37 @@ where
     /// ```
     pub fn get_ref(&self) -> &W {
         &self.inner
+    }
+
+    /// Writes a [`Line`].
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use std::io;
+    /// use noodles_gff as gff;
+    /// use gff::line::Line;
+    ///
+    /// let mut writer = gff::Writer::new(Vec::new());
+    ///
+    /// let version = Line::Directive(gff::Directive::GffVersion(Default::default()));
+    /// writer.write_line(&version)?;
+    ///
+    /// let comment = Line::Comment(String::from("noodles"));
+    /// writer.write_line(&comment)?;
+    ///
+    /// let record = Line::Record(gff::Record::default());
+    /// writer.write_line(&record)?;
+    ///
+    /// let expected = b"##gff-version 3
+    /// #noodles
+    /// .\t.\t.\t1\t1\t.\t.\t.\t.
+    /// ";
+    ///
+    /// assert_eq!(&writer.get_ref()[..], &expected[..]);
+    /// # Ok::<(), io::Error>(())
+    pub fn write_line(&mut self, line: &Line) -> io::Result<()> {
+        writeln!(self.inner, "{}", line)
     }
 
     /// Writes a GFF directive.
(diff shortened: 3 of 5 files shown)


claymcleod:
I also debated not making [the inner field `pub`](https://github.com/zaeleus/noodles/pull/120/files#diff-9606392498eb1b760ef93c3aef70915c3841c46e7e1cdf7e11528f1543ce7fdfR11) and creating a `Comment::new` function instead that takes a `String`. Let me know if you like that approach better.

zaeleus:
> since this is not really an unconstrained `String`

What do you mean by this?

Adding `Writer::write_line` instead would be a much simpler approach to this.

claymcleod:
> > since this is not really an unconstrained `String`
> 
> What do you mean by this?
> 
> Adding `Writer::write_line` instead would be a much simpler approach to this.

I simply mean that comments required to be preceded by a `#`:

> Blank lines should be ignored by parsers and lines beginning with a single '#' are used for human-readable comments and can be ignored by parsers. End-of-line comments (comments preceded by # at the end of and on the same line as a feature or directive line) are not allowed.

In other words, you can't write _any_ `String` to the file and it still be a valid GFF comment.

By `Writer::write_line`, I'm assuming that would allow you to write any `String` you want to the file, right? If so, couldn't you construct an invalid GFF file by simply choosing to not include the preceding `#` (e.g., `writer.write_line("hello, world")`?

claymcleod:
Oh, I see what you're saying now. Yeah, you're right, that would be simpler. I can change that if that's the way you want to do it; up to you.

zaeleus:
> I can change that if that's the way you want to do it

Yes, please.

claymcleod:
How about something like this? I was unsure if we should rework the existing `write_directive` and `write_record` to use this underneath the hood, but I decided against it (mainly because it would require consuming a `Directive`/`Record`, and the method takes them as immutable reference).

noodles-gff/examples/gff_write.rs
```diff
@@ -4,6 +4,7 @@
 
 use std::io;
 
+use gff::Line;
 use noodles_gff as gff;
```
zaeleus, line 8:
```suggestion
use noodles_gff::{self as gff, Line};
```

noodles-gff/src/writer.rs
```diff
@@ -1,5 +1,7 @@
 use std::io::{self, Write};
 
+use crate::Line;
+
 use super::{Directive, Record};
```
zaeleus, line 5:
```suggestion
use super::{Directive, Line, Record};
```

zaeleus (changes requested):

noodles-gff/examples/gff_view.rs
```diff
@@ -13,9 +13,12 @@ fn main() -> io::Result<()> {
 
     let mut reader = File::open(src).map(BufReader::new).map(gff::Reader::new)?;
 
-    for result in reader.records() {
-        let record = result?;
-        println!("{}", record);
+    let stdout = io::stdout().lock();
+    let mut writer = gff::Writer::new(stdout);
```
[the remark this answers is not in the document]
claymcleod: Did not know this, thanks for the check. I'll test it out with the MSRV.

noodles-gff/src/writer.rs
claymcleod: Good catch, this is an editor setting for me. My fault

claymcleod:
I've applied your changes. Once you're happy with it, I will squash the changes and then you can merge.

Just for my own information,

- What's the benefit of having multiple different MSRV for these packages?
- How do you typically run the tests for these with a different MSRV? Here's what I tried and subsequently got back:

```
$ rustup toolchain install 1.57.0 && rustup default 1.57.0
$ cd noodles-gff
$ cargo test --all-features
error: failed to load manifest for workspace member `noodles/noodles`

Caused by:
  failed to parse manifest at `noodles/noodles/Cargo.toml`

Caused by:
  optional dependency features with `?` syntax are only allowed on the nightly channel and requires the `-Z weak-dep-features` flag on the command line
  Feature `async` had feature value `noodles-bam?/async`.
```

zaeleus:
> What's the benefit of having multiple different MSRV for these packages?

noodles is a collection of libraries. Users aren't required to use the noodles meta crate, which inherently has to use the highest Rust version of all the crates for its minimum Rust version (currently 1.62.0). If, e.g., noodles-gff was the only crate needed, it can be pulled into a project independently, which would maintain a minimum Rust version of 1.57.0.

> How do you typically run the tests for these with a different MSRV?

My method isn't ideal (maybe I'm missing an option in the documentation?), but it works. 😅 I disable the virtual workspace and test the crate(s) individually, e.g.,

```
$ rm Cargo.toml
$ cd noodles-gff
$ rustup toolchain install 1.57.0
$ cargo +1.57.0 test
$ cd ..
$ git restore Cargo.toml
```

(Note that rustup creates a shim that allows `cargo` to optionally use a selected toolchain.)

[cargo-msrv](https://github.com/foresterre/cargo-msrv) is also useful to automate finding the minimum Rust version.

```
$ cargo msrv --min 2021
  [Meta]  cargo-msrv 0.15.1 (4335327)

Compatibility Check #1: Rust 1.60.0
    [OK]  Is compatible

Compatibility Check #2: Rust 1.58.1
    [OK]  Is compatible

Compatibility Check #3: Rust 1.57.0
    [OK]  Is compatible

Compatibility Check #4: Rust 1.56.1
  [FAIL]  Is incompatible
╭──────────────────────────────────────────────────────────────────────────────╮
│ error: package `noodles-core v0.9.0 (/tmp/noodles/noodles-core)` cannot be b │
│ uilt because it requires rustc 1.57.0 or newer, while the currently active r │
│ ustc version is 1.56.1                                                       │
╰──────────────────────────────────────────────────────────────────────────────╯

Result:
 Considered (min … max):   Rust 1.56.0 … Rust 1.64.0
 Search method:            bisect
 MSRV:                     1.57.0
 Target:                   x86_64-unknown-linux-gnu
```

In this case, the Rust version for noodles-gff must be 1.57.0 because it depends on noodles-core, which has a Rust version of 1.57.0.

claymcleod:
Got it, thanks for the explanation! I really was scratching my head as to the best way to test this.

All the tests pass with 1.57.0 and I've squashed the changes, so we should be good to merge if you're happy with it.

PR #127, opened 2022-10-31
gtf/writer: Adds `write_line` method to writer

claymcleod:
Adds a `write_line` method in the same manner as [`noodles/gtf`](https://github.com/zaeleus/noodles/commit/53b43c73e0b904778651bea02a5a529ff209688d).


zaeleus (approved):
Thanks!

PR #129, opened 2022-11-04
bgzf/reader: Change linear search to binary search

lu6huimin:
The index is sorted, we could do binary search ([`partition_point`](https://doc.rust-lang.org/core/primitive.slice.html#method.partition_point) method) on it.


zaeleus (approved):
Good catch, thanks!

PR #148, opened 2023-02-25
bgzf/gzi/reader: wrap gzi Reader in a BufReader

### LINKED ISSUE #143: Add BufRead option for gzi?
I was doing some performance testing for `noodles_bgzf` vs bgzip and noticed some unexpected slowdowns when monitoring syscalls with `strace`:

Baseline performance for seek operation seems comparable on a medium size (~175M) file:
```sh
$ time bgzf_seek_uncompressed contigs.fasta.gz 112093709 1149
real	0m0.022s
user	0m0.007s
sys	0m0.008s
```

```sh
$ time bgzip -b 112093709 -s 1149 contigs.fa.gz
real	0m0.016s
user	0m0.004s
sys	0m0.003s
```

But when `strace` is used:
```sh
$ time strace -o >(cat >strace_noodles.log) -e trace=open,openat,close,read,write bgzf_seek_uncompressed contigs.fa.gz 112093709 1149
real	0m2.018s
user	0m0.115s
sys	0m1.120s
```
```sh
$ time strace -o >(cat >strace_bgzip.log) -e trace=open,openat,close,read,write bgzip -b 112093709 -s 1149 contigs.fa.gz
real	0m0.056s
user	0m0.003s
sys	0m0.032s
```

After some sleuthing, it turns out the number of `read` calls on `contigs.fa.gz.gzi` is significant:
```sh
$ sed -n '/open.*\.gzi/,/close/p' strace_noodles.log  | grep -c read
9960
$  sed -n '/open.*\.gzi/,/close/p' strace_bgzip.log  | grep -c read
20
```

I suspect this will start to have performance impacts on huge files. Inspired by this particular [blog post](https://era.co/blog/unbuffered-io-slows-rust-programs) might I suggest a convenience function for buffered reads, similar in spirit to [`gzi::read`](https://github.com/zaeleus/noodles/blob/master/noodles-bgzf/src/gzi.rs#L35-L41)

I put together an example [in this gist](https://gist.github.com/zdk123/720a7c490a4e8f067d5ae97fce1b9a63):

```sh
$ time strace -o >(cat >strace_unbuf.log) -e trace=open,openat,close,read,write bgzf_gzi_inspect contigs.fa.gz.gzi
real	0m10.239s
user	0m0.290s
sys	0m2.933s
$ sed -n '/open.*\.gzi/,/close/p' strace_unbuf.log  | grep -c read
9960
```
vs
```sh
$ time strace -o >(cat >strace_buf.log) -e trace=open,openat,close,read,write target/release/examples/bgzf_bufgzi_inspect contigs.fa.gz.gzi > tmp.txt
real	0m8.654s
user	0m0.151s
sys	0m1.503s
$ sed -n '/open.*\.gzi/,/close/p' strace_buf.log  | grep -c read
2
```

This is just an example, of course - I'm not suggesting that the entire GZI file should be read in one shot, but some buffer size comparable to bgzip could make sense.

I'm happy to contribute more code if anyone things this is a good idea.

zdk123:
Fixes #143. 

`strace` is recording 11 read operations now, rather than  9960 - which is more inline with `bgzip`.

### DIFF
--- a/noodles-bgzf/src/gzi.rs
+++ b/noodles-bgzf/src/gzi.rs
@@ -14,8 +14,8 @@ pub use self::reader::Reader;
 #[cfg(feature = "async")]
 pub use self::r#async::Reader as AsyncReader;
 
-use std::{fs::File, io, path::Path};
 pub use std::io::BufReader;
+use std::{fs::File, io, path::Path};
 
 /// A gzip index.
 pub type Index = Vec<(u64, u64)>;
@@ -37,8 +37,6 @@ pub fn read<P>(src: P) -> io::Result<Index>
 where
     P: AsRef<Path>,
 {
-    let mut reader = File::open(src).map(BufReader::new)
-                                    .map(Reader::new)?;
-
+    let mut reader = File::open(src).map(BufReader::new).map(Reader::new)?;
     reader.read_index()
 }
--- a/noodles-bgzf/src/gzi.rs
+++ b/noodles-bgzf/src/gzi.rs
@@ -14,8 +14,11 @@ pub use self::reader::Reader;
 #[cfg(feature = "async")]
 pub use self::r#async::Reader as AsyncReader;
 
-pub use std::io::BufReader;
-use std::{fs::File, io, path::Path};
+use std::{
+    fs::File,
+    io::{self, BufReader},
+    path::Path,
+};
 
 /// A gzip index.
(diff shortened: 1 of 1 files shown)


zaeleus:
Great, thanks!

zdk123:
Thanks for the formatting changes

PR #149, opened 2023-03-01
util: Variant reader

malthesr:
I needed something like the `noodles_util::alignment` module, but for the common variant call formats (corresponding to the various `--output-type` options  in `bcftools`). I was just playing around to see if this could be easily done in the same spirit as for the alignment module, which seems to be the case.

This PR contains the reader only. If you're happy to accept something like this, I can have a look at the corresponding writer in a follow-up PR (or here, if you prefer). Otherwise, feel free to close and I'll just pull out what I need for private use.

### DIFF
--- a/noodles-util/src/variant/format.rs
+++ b/noodles-util/src/variant/format.rs
@@ -12,6 +12,4 @@ pub enum Format {
 pub enum Compression {
     /// BGZF compression.
     Bgzf,
-    /// No compression.
-    Uncompressed,
 }
--- a/noodles-util/src/variant/format.rs
+++ b/noodles-util/src/variant/format.rs
@@ -1,0 +1,17 @@
+/// A variant format.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Format {
+    /// Variant Call Format (VCF).
+    Vcf,
+    /// BCF.
+    Bcf,
+}
+
+/// A variant compression.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Compression {
+    /// BGZF compression.
+    Bgzf,
+    /// No compression.
+    Uncompressed,
+}
--- a/noodles-util/src/variant/reader/builder.rs
+++ b/noodles-util/src/variant/reader/builder.rs
@@ -15,7 +15,8 @@ use super::Reader;
 /// A variant reader builder.
 #[derive(Default)]
 pub struct Builder {
-    compression: Option<Compression>,
+    // None means infer on build; Some(None) means no compression explicitly set.
+    compression: Option<Option<Compression>>,
     format: Option<Format>,
 }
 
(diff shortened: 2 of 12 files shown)


noodles-util/src/variant/format.rs
```diff
@@ -0,0 +1,17 @@
...
+pub enum Format {
+    /// Variant Call Format (VCF).
+    Vcf,
+    /// BCF.
+    Bcf,
+}
+
+/// A variant compression.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Compression {
```
zaeleus, line 12: In noodles-cram, `BlockContentEncoderMap` encoders are `Option<Encoder>`, where `None` represents no compression. Should we take the same approach here, i.e., removing `Uncompressed` and having compression be an `Option<Compression>`?

noodles-util/src/variant/reader/builder.rs
```diff
@@ -0,0 +1,161 @@
...
+        let inner: Box<dyn VariantReader<_>> = match (format, compression) {
+            (Format::Vcf, Compression::Uncompressed) => {
+                let inner: Box<dyn BufRead> = Box::new(BufReader::new(reader));
+                Box::new(vcf::Reader::new(inner))
+            }
+            (Format::Vcf, Compression::Bgzf) => {
+                let inner: Box<dyn BufRead> = Box::new(bgzf::Reader::new(reader));
+                Box::new(vcf::Reader::new(inner))
+            }
+            (Format::Bcf, Compression::Uncompressed) => {
```
zaeleus, line 112: Fun fact: A raw BCF is actually invalid; the format is inherently BGZF-compressed. However, some libraries (e.g., HTSJDK) write raw BCF, and due to the output's prevalence, the VCF 4.4 spec now requires read support for it.

zaeleus (commented):
This is great work, thanks! I had this on my todo list but never got around to it. 😅

> I can have a look at the corresponding writer in a follow-up PR

Yes, if you'd like, it's tracked in #150. Please create a new PR for it.

noodles-util/src/variant/format.rs
malthesr: That's probably sensible! I've made the corresponding change.

noodles-util/src/variant/reader/builder.rs
malthesr: Huh, I didn't know that - I'm too used to writing uncompressed BCF using `bcftools`.  Thanks for the tip!

Should we keep the support for uncompressed BCF in anticipation of VCF 4.4, or would you prefer that I remove it in the interrim?

malthesr:
Thanks for the speedy review!

I'll have a go at #150 soon to complement the reader.

noodles-util/src/variant/reader/builder.rs
zaeleus: If you need uncompressed BCF, the spec recommends "files should be still compressed with BGZF but with compression 0" (§ 6.5 "BCF2 block gzip and indexing" (2022-08-22)).

Yes, keep it for interop and VCF 4.4 compliance.

zaeleus (approved):

PR #152, opened 2023-03-04
util: Variant writer

### LINKED ISSUE #150: util/variant: Add writer
Add a high-level writer that abstracts writing VCF records in VCF or BCF. It should add a `VariantWriter` trait and be similar to how `util::alignment::Writer` is implemented.

malthesr:
Closes #150 

The followup to #149, adding the `VariantWriter` trait in `noodles-vcf` and the `variant::Writer` in `noodles-util`.

As agreed in #150, I've left aside the issue of `StringMaps` in the impl for `VariantWriter` for now.

Mostly, this follows the implementation of `AlignmentWriter` and `alignment::Writer`, but I had to make a few choices about the compression behaviour for the BCF format that may warrant a bit of attention. Specifically,
- Specifying `None` for compression for BCF leads to a BGZF-compressed file using compression level 0.
- The default compression for BCF (and VCF) is `None`.
- The path extension `.bcf` leads to `None` for compression, unlike `.bcf.(gz|bgz)`.
As you mentioned in the previous PR, BCF should technically always be compressed, so I see how this could be debatable. Let me know if you have different preferences.

### DIFF
--- a/noodles-util/src/variant/writer/builder.rs
+++ b/noodles-util/src/variant/writer/builder.rs
@@ -102,18 +102,17 @@ impl Builder {
         W: Write + 'static,
     {
         let format = self.format.unwrap_or(Format::Vcf);
-        let compression = self.compression.unwrap_or(None);
+        let compression = self.compression.unwrap_or(match format {
+            Format::Vcf => None,
+            Format::Bcf => Some(Compression::Bgzf),
+        });
 
         let inner: Box<dyn vcf::VariantWriter> = match (format, compression) {
             (Format::Vcf, None) => Box::new(vcf::Writer::new(writer)),
             (Format::Vcf, Some(Compression::Bgzf)) => {
                 Box::new(vcf::Writer::new(bgzf::Writer::new(writer)))
             }
-            (Format::Bcf, None) => Box::new(bcf::Writer::from(
-                bgzf::writer::Builder::default()
-                    .set_compression_level(bgzf::writer::CompressionLevel::none())
-                    .build_with_writer(writer),
-            )),
+            (Format::Bcf, None) => Box::new(bcf::Writer::from(writer)),
             (Format::Bcf, Some(Compression::Bgzf)) => Box::new(bcf::Writer::new(writer)),
         };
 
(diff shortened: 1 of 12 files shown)


noodles-util/src/variant/writer/builder.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        let compression = self
+            .compression
+            .get_or_insert_with(|| detect_compression_from_path_extension(src));
+
+        if self.format.is_none() {
+            self.format = detect_format_from_path_extension(src, *compression);
+        }
+
+        let file = File::create(src)?;
```
malthesr, line 80: My instinct would have been to wrap this in a `io::BufWriter`, but I didn't to mirror the `alignment::Writer`.

zaeleus (changes requested):
> Specifying `None` for compression for BCF leads to a BGZF-compressed file using compression level 0.

`Some(None)` should write a raw BCF. Even though I think BCF should not be written raw, overriding this is confusing.

> The default compression for BCF (and VCF) is `None`.

The default compression for VCF should be uncompressed; and BCF, BGZF.

> The path extension `.bcf` leads to `None` for compression, unlike .`bcf.(gz|bgz)`.

As above, `bcf` should default to BGZF compression, not raw. Is `bcf.gz` a common suffix?

malthesr:
I'll get to this soon, just a few clarifications:

> As above, `bcf` should default to BGZF compression, not raw. 

That does make sense. Just to be clear, that means we're saying uncompressed BCF cannot be inferred from any path extension?

> Is `bcf.gz` a common suffix?

In the wild, I usually see `.bcf` to mean raw BCF, and `.bcf.gz` means compressed, so I see `.bcf.gz` more than `.bcf` but I can't say whether that's representative. Hopefully not, given how it's meant to be used. If we're never inferring uncompressed BCF from paths, then I think it makes sense to drop any support for `.bcf.gz`, so that the only special extensions are `.vcf`, `.vcf.{gz,bgz}`, and `bcf`.

zaeleus:
> Just to be clear, that means we're saying uncompressed BCF cannot be inferred from any path extension?

Correct.

> In the wild, I usually see `.bcf` to mean raw BCF, and `.bcf.gz` means compressed, so I see `.bcf.gz` more than `.bcf` but I can't say whether that's representative.

I'm curious what tooling is doing this. E.g., bcftool's inference doesn't support the `bcf.gz` suffix.

```
$ bcftools --version | head -1
bcftools 1.17
$ bcftools view --output simple.bcf simple.vcf
$ file simple.bcf
simple.bcf: Blocked GNU Zip Format (BGZF; gzip compatible), block length 1031
$ bcftools view --output simple.bcf.gz simple.vcf
$ file simple.bcf.gz
simple.bcf.gz: Variant Call Format (VCF) version 4.3, ASCII text
```

malthesr:
I'm not aware of any particular tooling doing it, I just see it as a convention to distinguish between compressed and uncompressed when piping output with `bcftools`. I think it arises because it somehow seems natural to distinguish between the four output modes `-O {v,z,u,b}` by path extension, and most people (like myself) are not aware that raw BCF is unofficial.

Either way, dropping any support `bcf.gz` seems to be the consistent way forward, and consumers can use the builder API to specify otherwise as needed.

zaeleus (approved):
Great, thanks for working on this!

PR #177, opened 2023-06-12
[bam] makes sequence writer no-op for empty sequence.

### LINKED ISSUE #176: How to (async) write secondary BAM records
Hello, 

Firstly, I'm really excited about using this library. My question is probably of the "holding it wrong" variety. I am investigating using the bam::AsyncReader and bam::AsyncWriter, when I encounter a secondary/reverse-complement secondary alignment I get this error from [write_record](https://docs.rs/noodles-bam/latest/noodles_bam/struct.AsyncWriter.html#method.write_record): `read length-sequence length mismatch`. I couldn't tell from the documentation if there is something I need to do in order to write these records. Thanks for the help. 

simplified code example
```rust
async fn main() -> anyhow::Result<()> {
    // setup..

    let mut reader = File::open(&cli.in_bam).await.map(bam::AsyncReader::new)?;
    let header = reader
        .read_header()
        .await?
        .parse()?;

    let _reference_sequences = reader
        .read_reference_sequences()
        .await?;

    let mut writer = File::create(&cli.out_bam)
        .await
        .map(|fh| bam::AsyncWriter::new(fh))?;
    writer
        .write_header(&header)
        .await?;
    writer
        .write_reference_sequences(header.reference_sequences())
        .await?;

    // 
    // ..snip...
    // 

    let mut records = reader.records(&header);
    while let Some(result) = records.next().await {
        match result {
            Ok(mut record) => {
                record.data_mut().insert(tag, tag_value.clone());
                let ok = writer.write_record(&header, &record).await;
                match ok {
                    Ok(_) => pb.inc(1),
                    Err(e) => {
                        let record_name = record.read_name().map(|n| n.to_string());
                        println!(
                            "there was an err {} with {:?}, flags {:?}",
                            e.to_string(),
                            record_name,
                            record.flags()
                        );
                    }
                }
            }
            Err(_) => erred += 1,
        }
    }
    writer.shutdown().await?;
    Ok(())
}

```

ArtRand:
Fixes #176 .

### DIFF
--- a/noodles-bam/src/record/codec/encoder/sequence.rs
+++ b/noodles-bam/src/record/codec/encoder/sequence.rs
@@ -7,6 +7,10 @@ pub fn put_sequence<B>(dst: &mut B, read_length: usize, sequence: &Sequence) -> 
 where
     B: BufMut,
 {
+    if sequence.is_empty() {
+        return Ok(());
+    }
+
     // § 1.4.10 "`SEQ`" (2022-08-22): "If not a '*', the length of the sequence must equal the sum
     // of lengths of `M`/`I`/`S`/`=`/`X` operations in `CIGAR`."
     if read_length > 0 && sequence.len() != read_length {
--- a/noodles-bam/src/record/codec/encoder/sequence.rs
+++ b/noodles-bam/src/record/codec/encoder/sequence.rs
@@ -80,6 +80,10 @@ mod tests {
         t(&mut buf, &"ACGT".parse()?, &[0x12, 0x48])?;
 
         buf.clear();
+        put_sequence(&mut buf, 2, &Sequence::default())?;
+        assert!(buf.is_empty());
+
+        buf.clear();
         let sequence = "A".parse()?;
         assert!(matches!(
             put_sequence(&mut buf, 2, &sequence),
--- a/noodles-bam/CHANGELOG.md
+++ b/noodles-bam/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+  * bam/record/codec/encoder/sequence: Fix writing an empty sequence when the
+    read length is nonempty ([#176]).
+
+[#176]: https://github.com/zaeleus/noodles/issues/176
+
 ## 0.35.0 - 2023-06-08
 
 ### Removed


noodles-bam/src/record/codec/encoder/sequence.rs
```diff
@@ -7,26 +7,27 @@ pub fn put_sequence<B>(dst: &mut B, read_length: usize, sequence: &Sequence) ->
 where
     B: BufMut,
 {
-    // § 1.4.10 "`SEQ`" (2022-08-22): "If not a '*', the length of the sequence must equal the sum
-    // of lengths of `M`/`I`/`S`/`=`/`X` operations in `CIGAR`."
-    if read_length > 0 && sequence.len() != read_length {
-        return Err(io::Error::new(
+    if sequence.is_empty() {
+        Ok(())
```
zaeleus, line 11: No need to change the shape of the entire function; this just needs to return early.

noodles-bam/src/record/codec/encoder/sequence.rs
```diff
@@ -7,26 +7,27 @@ pub fn put_sequence<B>(dst: &mut B, read_length: usize, sequence: &Sequence) ->
...
-    while let Some(l) = bases.next() {
-        // § 4.2.3 "SEQ and QUAL encoding" (2021-06-03): "When `l_seq` is odd the bottom 4 bits of
-        // the last byte are undefined, but we recommend writing these as zero."
-        let r = bases.next().unwrap_or(Base::Eq);
-        let b = encode_base(l) << 4 | encode_base(r);
-        dst.put_u8(b);
+            "read length and sequence length mismatch",
+        ))
+    } else {
+        let mut bases = sequence.as_ref().iter();
```
zaeleus, line 20: Why was the [`Copied` adapter](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.copied) removed from this iterator? It removes the need to deref the values below.

zaeleus (changes requested):

noodles-bam/src/record/codec/encoder/sequence.rs
ArtRand: Yup, no problem.

noodles-bam/src/record/codec/encoder/sequence.rs
ArtRand: Just a silly preference of mine, changed it back.

PR #199, opened 2023-08-31
feat: add lazy record reader to vcf async reader

tshauck:
Hi -- I noticed you recently added a lazy record for VCF reader. This has been nice in cases where I'm using the non-async reader. So, hoping to have a similar speed up in the async case, I'd like to add `read_lazy_record` to the async vcf reader. This is mostly a copy of the sync implementation, just w/ awaits and different trait bounds.

Thanks!


zaeleus (approved):
Great, thanks!

zaeleus:
This is now available in noodles 0.50.0 / noodles-vcf 0.38.0.

PR #206, opened 2023-10-01
Add reserve method to Cigar for use in `get_cigar`

tshauck:
Hi,

I'd like to propose a PR that I think could improve BAM cigar parsing performance. I noticed while profiling a tool of mine that calls to `get_cigar` may have do n allocations based on the number of cigar ops. Looking at the `get_cigar` function, it already takes the number of ops, so this PR adds a `reserve` method on `Cigar` so that `get_cigar` can make use of it.

I did a little benchmark via criterion that seems to indicate reduced runtime and reduced runtime variance.

image

I put the benchmark code in a separate branch here: https://github.com/tshauck/noodles/tree/add-reserve-to-cigar-decode-benches/target/criterion (with some changes to `noodles-bam` too).

### DIFF
--- a/noodles-bam/src/record/codec/decoder/cigar.rs
+++ b/noodles-bam/src/record/codec/decoder/cigar.rs
@@ -60,6 +60,7 @@ where
     }
 
     cigar.clear();
+    cigar.reserve(n_cigar_op);
 
     for _ in 0..n_cigar_op {
         let op = decode_op(src.get_u32_le()).map_err(DecodeError::InvalidOp)?;
--- a/noodles-sam/src/record/cigar.rs
+++ b/noodles-sam/src/record/cigar.rs
@@ -35,7 +35,7 @@ impl Cigar {
     /// # Examples
     ///
     /// ```no_run
-    /// use noodles_sam::record::{cigar::{op::Kind, Op}, Cigar};
+    /// use noodles_sam::record::Cigar;
     ///
     /// let mut cigar: Cigar = Cigar::default();
     /// cigar.reserve(1);
--- a/noodles-sam/src/record/cigar.rs
+++ b/noodles-sam/src/record/cigar.rs
@@ -30,6 +30,20 @@ impl Cigar {
         self.0.clear();
     }
 
+    /// Reserves capacity for additional operations.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use noodles_sam::record::{cigar::{op::Kind, Op}, Cigar};
+    ///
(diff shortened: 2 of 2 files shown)


zaeleus:
While this performance difference can be observed in how your benchmarks are defined, i.e., creating a new buffer for each iteration; in practice, there shouldn't be a difference with the current BAM decoder. The CIGAR buffer in a `sam::alignment::Record` is reused for each record, and `Vec::clear` does not affect its internal capacity. (Also note that `Vec::reserve` is for an _additional_ number of entries, not for the expected count unless the capacity is already 0.)

> I noticed while profiling a tool of mine that calls to `get_cigar` may have do n allocations based on the number of cigar ops.

Can you share this report? When a `Vec` is empty, the first push allocates for a minimum of 4 items, and when at capacity, the size of the `Vec` doubles. E.g., there will be an allocation at item 1 (capacity 4), 5 (capacity 8), 9 (capacity 16), etc.

zaeleus:
Ah, I see. Since this only affects the conversion use case, can we use the ops iterator instead of the decoder in the implementation of `lazy::record::Cigar::try_from`, i.e., call `self.iter().collect()`? The iterator has the correct size hint set, which preallocates the resulting `Vec`.

tshauck:
Cool, thanks for the feedback. Please let me know if update is what you had in mind.

noodles-bam/src/lazy/record/cigar.rs
```diff
@@ -49,15 +49,8 @@ impl<'a> TryFrom<Cigar<'a>> for sam::record::Cigar {
...
-
-        let mut src = bam_cigar.0;
-        let mut cigar = Self::default();
-        let op_count = bam_cigar.len();
-        get_cigar(&mut src, &mut cigar, op_count)
-            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
-
-        Ok(cigar)
+        let ops = bam_cigar.iter().collect::<io::Result<Vec<_>>>()?;
+        Ok(Self::from_iter(ops))
```
zaeleus, line 53:
```suggestion
        bam_cigar.iter().collect()
```

zaeleus (changes requested):

noodles-bam/src/lazy/record/cigar.rs
tshauck: lol, much nicer

zaeleus (approved):

zaeleus:
Thanks! (And sorry for the slow responses; I just got back from vacation.)

tshauck:
All good -- appreciate the education/patience

PR #223, opened 2024-01-22
fix: fix range for INFO for lazy vcf record bounds

tshauck:
Hi,

I think there's an issue with `info_range` for the lazy vcf record in that currently it's actually the range for `FILTERS`, so rather than being the info in the record, the filter info is duplicated.

So this PR updates that `info_range` method and adds tests that check the bounds defined in the object come out correctly via the methods.

Thanks,


zaeleus (approved):

zaeleus:
Yes, thanks for catching!

PR #245, opened 2024-03-13
Add new BAM example for full PG entry

jamestwebber:
This is an example script that adds a full PG tag to the header of a BAM file (`sample.bam` is the output of `bam_write`):

```
@HD	VN:1.6
@PG	ID:noodles-bam
@PG	ID:bam_add_tag	PN:bam_add_tag	PP:noodles-bam	VN:0.57.0	CL:target/release/examples/bam_add_tag_to_header sample.bam
@PG	ID:samtools	PN:samtools	PP:bam_add_tag	VN:1.19	CL:samtools view --with-header
@CO	an example BAM written by noodles-bam
*	4	*	0	255	*	*	0	0	*	*
```

I spent some time yesterday figuring out how to do this, and I figured the example might be useful. I'm also opening the PR to see if there are any improvements you can see.

The name could be more clear, perhaps `bam_add_pg_entry` would be better. People might expect "tag" to refer to the reads.

Constructing the `PN`, `PP`, `VN`, and `CL` tags seems cumbersome. Maybe these should be built-in tags (I believe they're part of the spec). Or if there's a way to construct them more cleanly I'd be happy to use that.

### DIFF
--- a/noodles-bam/examples/bam_add_tag_to_header.rs
+++ b/noodles-bam/examples/bam_add_tag_to_header.rs
@@ -8,7 +8,7 @@ use std::{env, io};
 
 use noodles_bam as bam;
 use noodles_sam::header::record::value::{
-    map::{Program, Tag},
+    map::{program::tag, Program},
     Map,
 };
 use noodles_sam::Header;
@@ -17,37 +17,20 @@ fn add_pg(mut header: Header) -> Header {
     const NAME: &str = "bam_add_tag_to_header";
     const VERSION: &str = env!("CARGO_PKG_VERSION");
 
-    // standard PG tags PN, VN, PP, CL
-    let pn = match Tag::try_from([b'P', b'N']) {
-        Ok(Tag::Other(tag)) => tag,
-        _ => unreachable!(),
-    };
-    let vn = match Tag::try_from([b'V', b'N']) {
-        Ok(Tag::Other(tag)) => tag,
-        _ => unreachable!(),
-    };
-    let pp = match Tag::try_from([b'P', b'P']) {
-        Ok(Tag::Other(tag)) => tag,
-        _ => unreachable!(),
-    };
-    let cl = match Tag::try_from([b'C', b'L']) {
-        Ok(Tag::Other(tag)) => tag,
-        _ => unreachable!(),
-    };
     // the command-line to insert into the CL tag
     let cmd_str = env::args().collect::<Vec<_>>().join(" ");
 
-    let program = Map::<Program>::builder().insert(pn, NAME);
+    let program = Map::<Program>::builder().insert(tag::NAME, NAME);
 
     let program = if let Some(last_pg) = header.programs().keys().last() {
-        program.insert(pp, last_pg.clone())
+        program.insert(tag::PREVIOUS_PROGRAM_ID, last_pg.clone())
     } else {
         program
     };
 
     let program = program
-        .insert(vn, VERSION)
-        .insert(cl, cmd_str)
+        .insert(tag::VERSION, VERSION)
+        .insert(tag::COMMAND_LINE, cmd_str)
         .build()
         .unwrap();
     header.programs_mut().insert(NAME.into(), program);
--- a/noodles-bam/examples/bam_add_tag_to_header.rs
+++ b/noodles-bam/examples/bam_add_tag_to_header.rs
@@ -1,0 +1,73 @@
+//! Add a PG tag to the SAM header of a BAM file.
+//!
+//! This is a demonstration of how to construct a full PG entry.
+//!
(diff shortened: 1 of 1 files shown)


noodles-bam/examples/bam_add_tag_to_header.rs
```diff
@@ -0,0 +1,73 @@
...
+    let pp = match Tag::try_from([b'P', b'P']) {
+        Ok(Tag::Other(tag)) => tag,
+        _ => unreachable!(),
+    };
+    let cl = match Tag::try_from([b'C', b'L']) {
+        Ok(Tag::Other(tag)) => tag,
+        _ => unreachable!(),
+    };
+    // the command-line to insert into the CL tag
+    let cmd_str = env::args().collect::<Vec<_>>().join(" ");
```
zaeleus, line 38: Extra care needs to be taken for arguments containing tab characters.

noodles-bam/examples/bam_add_tag_to_header.rs
```diff
@@ -0,0 +1,73 @@
...
+        _ => unreachable!(),
+    };
+    let pp = match Tag::try_from([b'P', b'P']) {
+        Ok(Tag::Other(tag)) => tag,
+        _ => unreachable!(),
+    };
+    let cl = match Tag::try_from([b'C', b'L']) {
+        Ok(Tag::Other(tag)) => tag,
+        _ => unreachable!(),
+    };
```
zaeleus, line 36: Common tags are available as [constants in the `map::program::tag` module](https://docs.rs/noodles/0.66.0/noodles/sam/header/record/value/map/program/tag/index.html#constants).

noodles-bam/examples/bam_add_tag_to_header.rs
```diff
@@ -0,0 +1,73 @@
...
+    // the command-line to insert into the CL tag
+    let cmd_str = env::args().collect::<Vec<_>>().join(" ");
+
+    let program = Map::<Program>::builder().insert(pn, NAME);
+
+    let program = if let Some(last_pg) = header.programs().keys().last() {
+        program.insert(pp, last_pg.clone())
+    } else {
+        program
+    };
```
zaeleus, line 46: Header records are not guaranteed to be ordered by recency. This would be incorrect for, e.g.,

```
@PG	ID:pg1	PP:pg0
@PG	ID:pg0
```

zaeleus (commented):
I like the intentions of this, but handling this properly is more complex than what should be shown in an example. I'd like to first work on program record chaining before adding something like this. What you have now can certainly be used as a simple workaround.

noodles-bam/examples/bam_add_tag_to_header.rs
jamestwebber: Somehow managed to miss those!

jamestwebber:
I modified this to use the defined tags in `map::program::tag`, but it's okay to close this for now until you feel the API is ready. This PR was just a sneaky way for me to ask about those tags. 😂

zaeleus:
The SAM header programs are now wrapped by [`sam::header::Programs`](https://github.com/zaeleus/noodles/blob/f23008df69ee5c1a84b662f502815354d638d6b8/noodles-sam/src/header/programs.rs#L11-L40), which includes [`Programs::add`](https://github.com/zaeleus/noodles/blob/f23008df69ee5c1a84b662f502815354d638d6b8/noodles-sam/src/header/programs.rs#L71). This handles either creating a new program chain or adding to existing ones, e.g.,

https://github.com/zaeleus/noodles/blob/f23008df69ee5c1a84b662f502815354d638d6b8/noodles-sam/src/header/programs.rs#L52-L69

Note how pg0 is the root of the chain and pg1 links to pg0 via the previous program ID (`PP`) tag. It also covers more convoluted cases, e.g.,

https://github.com/zaeleus/noodles/blob/f23008df69ee5c1a84b662f502815354d638d6b8/noodles-sam/src/header/programs.rs#L229-L291

I added creating the self program record in the reheader examples, e.g.,

https://github.com/zaeleus/noodles/blob/f23008df69ee5c1a84b662f502815354d638d6b8/noodles-bam/examples/bam_reheader.rs#L15-L23

```text
$ cargo --quiet run --example bam_write | cargo --quiet run --example bam_reheader /dev/stdin | samtools view --no-PG --with-header
@HD	VN:1.6
@PG	ID:noodles-bam
@PG	ID:noodles	PN:bam_reheader	VN:0.59.0	CL:target/debug/examples/bam_reheader /dev/stdin	PP:noodles-bam
@CO	an example BAM written by noodles-bam
@CO	a comment added by noodles-bam
*	4	*	0	255	*	*	0	0	*	*
```

Thanks for initial exploration of this issue!

PR #270, opened 2024-06-14
bed/record: Return ParseError::MissingScore instead of ParseError::Mi…

ghuls:
…ssingName for score BED column


zaeleus:
Thanks!

PR #290, opened 2024-08-21
adds example for how to use the multi-threaded bgzf reader with bams

bwlang:
I found this q/a helpful, but didn't find it until I had struggled for a while... does this help to raise the profile a bit?
Or maybe there is a good reason not to use this approach?

(from https://github.com/zaeleus/noodles/discussions/280)


zaeleus:
Thanks for the suggestion. I added more generalized examples in c5fa91351c2015887a390a6f454232b24dde6dc7 to show how to swap the BGZF decoder.

Issue #286, opened 2024-07-23
Unified async alignment reader/writer


mbhall88:
I use the `noodles_util::alignment` readers and writers a lot. How easy would it be to implement an aync version of this transparent reader/writer?

If this is something you think would be reasonably straightforward I would be willing to draft a PR with some guidance on where to start.

zaeleus:
> If this is something you think would be reasonably straightforward I would be willing to draft a PR with some guidance on where to start.

Yes, give it a go! It's a loaded issue, consisting of four parts:

  * [ ] the async alignment reader,
  * [ ] its builder,
  * [ ] the async alignment writer,
  * [ ] and also its builder.

I recommend focusing on each individually, submitting a patch for review per part. It is a fair amount of work/code, but don't feel obligated to complete it all.

For the first part, I normalized the alignment format async reader methods, so you shouldn't have any problems outside of util. I also added an async variant format reader and builder to use as a reference. 

Here are some instructions on how to proceed with the reader:

1. Add the alignment format dependencies (SAM, BAM, and CRAM) to the `async` feature in `noodles-util/Cargo.toml`. These use [the optional dependency feature syntax](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features).

    https://github.com/zaeleus/noodles/blob/782d892196a30e856560e060c390ee20de774859/noodles-util/Cargo.toml#L23-L28

3. Add the modules.

    * `noodles-util/src/alignment/async.rs` ([example](https://github.com/zaeleus/noodles/blob/782d892196a30e856560e060c390ee20de774859/noodles-util/src/variant/async.rs))
    * `noodles-util/src/alignment/async/io.rs` ([example](https://github.com/zaeleus/noodles/blob/782d892196a30e856560e060c390ee20de774859/noodles-util/src/variant/async/io.rs))
    * `noodles-util/src/alignment/async/io/reader.rs` ([example](https://github.com/zaeleus/noodles/blob/master/noodles-util/src/variant/async/io/reader.rs))

4. Implement `Reader<R>`. This will not use the I/O trait approach like the blocking version. Since well know all the format variants, use an enum instead. Feel free to start with the following scaffold.

```rust
//! Async alignment reader.

/// An async alignment reader.
pub enum Reader<R> {
    /// SAM.
    Sam(sam::r#async::io::Reader<R>),
    /// BAM.
    Bam(bam::r#async::io::Reader<R>),
    /// CRAM.
[...]

PR #292, opened 2024-08-22
Add async alignment writer

mbhall88:
See #286 for more details. 

Checklist

- [ ] async Writer
- [x] Builder
- [x] Documentation examples
- [ ] `examples/` entry

Currently, I cannot compile the async Writer due to

```
error[E0308]: mismatched types
   --> noodles-util/src/alignment/async/io/writer.rs:80:63
    |
80  |             Self::Cram(writer) => writer.write_record(header, record).await,
    |                                          ------------         ^^^^^^ expected `Record`, found `&dyn Record`
    |                                          |
    |                                          arguments to this method are incorrect
    |
    = note: expected struct `noodles_cram::Record`
            found reference `&dyn noodles_sam::alignment::Record`
note: method defined here
   --> /Users/michael/Projects/noodles/noodles-cram/src/async/io/writer.rs:181:18
    |
181 |     pub async fn write_record(
    |                  ^^^^^^^^^^^^
```

Which is coming from the fact that `cram::async::io::Writer::write_record` takes a `cram::Record` type rather than a `&dyn sam::alignment::Record` trait.

https://github.com/zaeleus/noodles/blob/174ad3c98092728a6be3feabb346619ebd14a5ad/noodles-cram/src/async/io/writer.rs#L184

Is that a change that should be made in a separate PR?

In addition, I also noticed some slight differences in function naming schemes between SAM/BAM and CRAM. For example, the SAM and BAM async `Writer`s have a `write_header` function, however, the reciprocal CRAM function is `write_file_header` - not sure if this was intended though. Also, the CRAM async builder has the function `build_with_writer` rather than the `build_from_writer`. I appreciate these are minor things, but thought I would raise them in case you want to move this library towards consistent names across modules/crates. I guess for backwards compatibility we could add the [`#[deprecated`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute) attribute if you want to introduce the consistent named version.

### DIFF
--- a/noodles-util/src/alignment/async/io.rs
+++ b/noodles-util/src/alignment/async/io.rs
@@ -1,5 +1,7 @@
 //! Async alignment format I/O.
 
 pub mod reader;
+pub mod writer;
 
 pub use self::reader::Reader;
+pub use self::writer::Writer;
--- a/noodles-util/src/alignment/async/io/writer.rs
+++ b/noodles-util/src/alignment/async/io/writer.rs
@@ -1,0 +1,83 @@
+//! Async alignment writer.
+
+mod builder;
+
+use noodles_bam as bam;
+use noodles_cram as cram;
+use noodles_sam as sam;
+use tokio::io::{self, AsyncWrite};
+
+pub use self::builder::Builder;
+
+/// An async alignment writer.
+pub enum Writer<W: tokio::io::AsyncWrite> {
+    /// SAM.
+    Sam(sam::r#async::io::Writer<W>),
+    /// BAM.
+    Bam(bam::r#async::io::Writer<W>),
+    /// CRAM.
+    Cram(cram::r#async::io::Writer<W>),
+}
+
+impl<W> Writer<W>
+where
+    W: AsyncWrite + Unpin,
+{
+    /// Writes a SAM header.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[tokio::main]
+    /// # async fn main() -> tokio::io::Result<()> {
+    /// use noodles_sam as sam;
+    /// use noodles_util::alignment::r#async::io::writer::Builder;
+    /// use tokio::io;
+    /// use tokio::io::AsyncWriteExt;
+    ///
+    /// let mut writer = Builder::default().build_from_writer(io::sink());
+    ///
+    /// let header = sam::Header::default();
+    /// writer.write_header(&header).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn write_header(&mut self, header: &sam::Header) -> io::Result<()> {
+        match self {
+            Self::Sam(writer) => writer.write_header(header).await,
+            Self::Bam(writer) => writer.write_header(header).await,
+            Self::Cram(writer) => writer.write_file_header(header).await,
+        }
+    }
+
+    /// Writes an alignment record.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[tokio::main]
+    /// # async fn main() -> tokio::io::Result<()> {
+    /// use noodles_sam as sam;
+    /// use noodles_util::alignment::r#async::io::writer::Builder;
+    /// use tokio::io;
+    /// use tokio::io::AsyncWriteExt;
+    ///
+    /// let mut writer = Builder::default().build_from_writer(io::sink());
+    ///
+    /// let record = sam::Record::default();
+    /// writer.write_record(&record).await?;
+    /// # Ok(())
+    /// # }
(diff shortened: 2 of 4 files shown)


zaeleus:
> ```
>     = note: expected struct `noodles_cram::Record`
>             found reference `&dyn noodles_sam::alignment::Record`
> ```
>
> Is that a change that should be made in a separate PR?

For now, convert `&dyn sam::alignment::Record` to a `cram::Record` using [`cram::Record::try_from_alignment_record`](https://docs.rs/noodles/0.79.0/noodles/cram/struct.Record.html#method.try_from_alignment_record). I can normalize the interfaces after this patch.

> In addition, I also noticed some slight differences in function naming schemes between SAM/BAM and CRAM. For example, the SAM and BAM async `Writer`s have a `write_header` function, however, the reciprocal CRAM function is `write_file_header` - not sure if this was intended though.

Yes, this is intentional. I/O is modeled after the physical structure for the format (see § 5 "[File structure](https://samtools.github.io/hts-specs/CRAMv3.pdf#section.5)" (2024-03-21) for a visual representation). The file header is named to differentiate itself from the many other headers in the format, e.g., the container header, the slice header, the block header, etc. I did add a convenience method to the blocking reader in 2da4eae513c3de8b7ecd64ff0246fa61c56a58bb, which simply needs to be synchronized with the async reader later. Use the currently available methods to write a default file definition, and use the SAM header to write the file header, e.g.,

https://github.com/zaeleus/noodles/blob/174ad3c98092728a6be3feabb346619ebd14a5ad/noodles-cram/examples/cram_write_async.rs#L73-L74

> Also, the CRAM async builder has the function `build_with_writer` rather than the `build_from_writer`.

Good catch; it's definitely a typo. Use the methods that are currently available. I can also fix this after this patch.

mbhall88:
Further to the above, when adding the async Writer example I realised I had missed adding `Writer::finish`. Again, the individual BAM and CRAM methods that are analogous are called `shutdown` instead of `finish`. I took the name finish to mirror the sync alignment API. Not sure if these too should be harmonised?

Aside from that, I *think* this PR is done and ready for review.

noodles-util/src/alignment/async/io/writer.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// let mut writer = alignment::r#async::io::writer::Builder::default()
+    ///     .set_format(Format::Sam)
+    ///     .build_from_writer(io::sink()).await?;
+    ///
+    /// let header = sam::Header::default();
+    /// writer.finish(&header).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn finish(&mut self, header: &sam::Header) -> io::Result<()> {
```
zaeleus, line 109: > Again, the individual BAM and CRAM methods that are analogous are called `shutdown` instead of `finish`. I took the name finish to mirror the sync alignment API. Not sure if these too should be harmonised?

This is intentional. Async I/O uses the finalization term `shutdown`, e.g., [`tokio::io::AsyncWriteExt::shutdown`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.shutdown). In blocking I/O, we typically rely on `drop` for implicit finalization. For explicit finalization, we just use the same terminology as used in common writers like [`flate2::write::GzEncoder::finish`](https://docs.rs/flate2/latest/flate2/write/struct.GzEncoder.html#method.finish).

Since this is async I/O, this method should use the name `shutdown` instead of `finish`.

noodles-util/src/alignment/async/io.rs
```diff
@@ -1,5 +1,7 @@
 //! Async alignment format I/O.
 
 pub mod reader;
+pub mod writer;
 
 pub use self::reader::Reader;
+pub use self::writer::Writer;
```
zaeleus, line 7:
```suggestion
pub use self::{reader::Reader, writer::Writer};
```

noodles-util/src/alignment/async/io/writer/builder.rs
```diff
@@ -0,0 +1,187 @@
...
+        self
+    }
+
+    /// Sets the reference sequence repository.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use noodles_fasta as fasta;
+    /// use noodles_util::alignment::{r#async::io::writer::Builder, io::Format};
```
zaeleus, line 62:
```suggestion
    /// use noodles_util::alignment::r#async::io::writer::Builder;
```

zaeleus (changes requested):

zaeleus (approved):
Looks good, thanks!

PR #323, opened 2025-01-29
make csi::io::indexed_records pub for access to ::record::Record

brentp:
see: https://github.com/zaeleus/noodles/discussions/322


zaeleus:
Thanks!

PR #331, opened 2025-03-03
Switch to more performant, 100% Rust zlib

brainstorm:
More information here: https://trifectatech.org/blog/zlib-rs-is-faster-than-c/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ bitflags = "2.0.2"
 bstr = { version = "1.9.0", default-features = false, features = ["std"] }
 byteorder = "1.2.3"
 bytes = "1.10.0"
-flate2 = "1.0.1"
+flate2 = { version = "1.1.0", features = ["zlib-rs"] }
 futures = { version = "0.3.15", default-features = false }
 indexmap = "2.0.0"
 lexical-core = "1.0.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ bitflags = "2.0.2"
 bstr = { version = "1.9.0", default-features = false, features = ["std"] }
 byteorder = "1.2.3"
 bytes = "1.10.0"
-flate2 = { version = "1.1.0", features = ["zlib-rs"] }
+flate2 = { version = "1.1.0", default-features = false, features = ["zlib-rs"] }
 futures = { version = "0.3.15", default-features = false }
 indexmap = "2.0.0"
 lexical-core = "1.0.0"


Cargo.toml
```diff
@@ -31,7 +31,7 @@ bitflags = "2.0.2"
 bstr = { version = "1.9.0", default-features = false, features = ["std"] }
 byteorder = "1.2.3"
 bytes = "1.10.0"
-flate2 = "1.0.1"
+flate2 = { version = "1.1.0", features = ["zlib-rs"] }
```
zaeleus, line 34:
```suggestion
flate2 = { version = "1.1.0", default-features = false, features = ["zlib-rs"] }
```
Otherwise, miniz_oxide is still compiled.

zaeleus (commented):
Thanks! I've also been following zlib-rs' progress and am impressed by its performance. Some things to note:

  1. Users can always enable the zlib-rs backend downstream, but if it were to be made the default in noodles, the selected backend would no longer be able to be reverted (or even changed to another), due to how features are additive in Cargo.

  2. [There are concerns](https://github.com/rust-lang/flate2-rs/issues/469#issuecomment-2687703783) over the amount of unsafe code used in zlib-rs, which [seems to have blocked it](https://github.com/rust-lang/flate2-rs/pull/470#pullrequestreview-2649894840) from eventually becoming the default backend for flate2.

  3. More of an observation, but it's quite weird how zlib-rs' integration in flate2 goes through the C API. Though I do understand the drop in simplicity and that its Rust API is considered unstable.

brainstorm:
Good issue reads, indeed! I've looked through the unsafe safety comments and it seems that they have been very diligent (ISRG seems to be auditing that code too).

Agreed that the amount of unsafe is surprising but most of it is needed (i.e to reach SIMD intrinsics and C FFI).

I don't have a strong opinion on this but I think it's worth a try: it can be reverted on future releases if it becomes problematic and shows good support of zlib-rs efforts (your crate is increasingly popular!).

zaeleus:
Yes, let's try it.

Measurements for posterity:

### Decode

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `bgzf_read_miniz_oxide` | 3.554 ± 0.012 | 3.544 | 3.576 | 2.09 ± 0.01 |
| `bgzf_read_zlib_rs` | 2.254 ± 0.006 | 2.247 | 2.263 | 1.32 ± 0.01 |
| `bgzf_read_libdeflate` | 1.704 ± 0.006 | 1.697 | 1.715 | 1.00 |

### Encode (uncompressed size: 1488620154 bytes; compression level 6)

| Command | Mean [s] | Min [s] | Max [s] | Relative | Compressed size [bytes] | Compression ratio |
|:---|---:|---:|---:|---:|---:|---:|
| `bgzf_write_miniz_oxide` | 51.214 ± 0.056 | 51.116 | 51.307 | 3.14 ± 0.02 | 366718357 | 24.6% |
| `bgzf_write_zlib_rs` | 21.761 ± 0.074 | 21.691 | 21.944 | 1.34 ± 0.01 | 366289772 | 24.6% |
| `bgzf_write_libdeflate` | 16.294 ± 0.085 | 16.163 | 16.406 | 1.00 | 367717478 | 24.7% |

PR #337, opened 2025-04-15
vcf/record/info: Avoid parsing skipped fields

QuartzLibrary:
I have an indexed vcf file, and my program was spending ~half the time (according to a quick flamegraph) parsing the info fields looking for `END` while querying it.
This just skips the parsing of the fields we don't care about on `get`. `variant_end` goes from ~45% to ~14%[1] the flamegraph for my usecase, not the most accurate assessment, but the change seems straightforward enough.

Thank you for the great libraries!

[1] It's very noisy, but it settled around here after some more extreme initial numbers.

### DIFF
--- a/noodles-vcf/src/record/info.rs
+++ b/noodles-vcf/src/record/info.rs
@@ -1,6 +1,6 @@
 mod field;
 
-use std::{io, iter};
+use std::io;
 
 use self::field::parse_field;
 use crate::{variant::record::info::field::Value, Header};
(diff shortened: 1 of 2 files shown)


noodles-vcf/src/record/info.rs
```diff
@@ -20,34 +20,33 @@ impl<'r> Info<'r> {
...
+
+        let field = self.iter_raw().find(|s| {
+            let Some(rest) = s.strip_prefix(key) else {
+                return false;
+            };
+            rest.is_empty() || rest.starts_with(SEPARATOR)
+        })?;
 
-        None
+        Some(parse_field(field, header).map(|(_, v)| v))
```
zaeleus, line 32: Nice optimization.

Instead of adding a raw iterator, we can simplify this by splitting on the delimiter and finding the key, e.g.,

```rust
const DELIMITER: char = ';';
let mut src = self.0.split(DELIMITER).find(|t| t.starts_with(key))?;
Some(parse_field(&mut src, header).map(|(_, v)| v))
```

QuartzLibrary: :+1: I wasn't sure how far on the DRY spectrum the style preference was, I'll remove the `iter_raw`.
Though the `rest.is_empty() || rest.starts_with(SEPARATOR)` is needed because otherwise `A` would match both of keys in `AB=1;A=1` otherwise.

PR #351, opened 2025-09-16
move from unmaintained xz2 to liblzma

xosxos:
Heads up. I couldn't compile noodles with the cram feature in a project due to it linking to older versions of lzma. I had to update xz2-rs to libzma-rs. 

All tests are passing

[https://github.com/portable-network-archive/liblzma-rs](https://github.com/portable-network-archive/liblzma-rs)

### DIFF
--- a/noodles-cram/Cargo.toml
+++ b/noodles-cram/Cargo.toml
@@ -26,7 +26,7 @@ noodles-bam = { path = "../noodles-bam", version = "0.83.0" }
 noodles-core = { path = "../noodles-core", version = "0.18.0" }
 noodles-fasta = { path = "../noodles-fasta", version = "0.56.0" }
 noodles-sam = { path = "../noodles-sam", version = "0.79.0" }
-xz2 = "0.1.6"
+liblzma = "0.3.6"
 
 async-compression = { version = "0.4.0", optional = true, features = ["gzip", "tokio"] }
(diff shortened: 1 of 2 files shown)


xosxos:
Well, I chose the last 3.x version because it made my stuff compile (I happen to be using [niffler](https://github.com/luizirber/niffler)). The 4.x version was only 6 months old, so I figured there could be compatibility issues with other packages as well.

That being said, I really don't know much about this, and xz was obviously targeted in the famous backdoor attempt (`liblzma-rs` version 3.2 actually [contains](https://github.com/Portable-Network-Archive/liblzma-rs/issues/106) the test files from xz/libzma C library that were used in the backdoor).

However, it's also not that good the `xz2` dependency is unmaintained. Although, `xz2` maintainer did note in [Jul 8, 2024](https://github.com/alexcrichton/xz2-rs/issues/128#issuecomment-2214139875) that he _might_ be available if major vulns or other issues come up.

zaeleus:
Ah, it's an unfortunate linking issue, which makes this a little more challenging to change. Migrating noodles to use liblzma would then not solve your issue, as niffler's direct dependency on it would still be behind. If your I/O is not LZMA-compressed, one workaround is to disable niffler's `lzma` feature.

As noted, xz2 is not actively maintained but not unmaintained. By default, it links to the system library and doesn't build and statically link to xz 5.2.

PR #364, opened 2025-12-29
feat(noodles-bam): expose record codec encode/decode functions

nh13:
## Summary

Expose the `decode` and `encode` functions in `noodles_bam::record::codec` as public API.

## Motivation

These functions enable direct byte-level manipulation of BAM records, which is useful for:
- Parallel BAM processing pipelines that decompress BGZF blocks independently
- Custom serialization workflows that need to encode/decode records without the 4-byte block_size prefix
- Performance-critical applications that bypass the standard reader/writer abstractions

## Changes

- Remove `#![doc(hidden)]` from the codec module
- Change `decode` function visibility from `pub(crate)` to `pub`
- Change `encode` function visibility from `pub(crate)` to `pub`
- Re-export `DecodeError` for proper error handling
- Add documentation for both functions

## API

```rust
/// Decodes a BAM record from raw bytes.
///
/// The input slice should contain the record data after the 4-byte block size.
/// The slice reference is advanced past the consumed bytes.
pub fn decode(src: &mut &[u8], record: &mut RecordBuf) -> Result<(), DecodeError>

/// Encodes a record as raw BAM record data.
///
/// The record is encoded without the leading 4-byte block size.
pub fn encode<R>(dst: &mut Vec<u8>, header: &Header, record: &R) -> io::Result<()>
where
    R: Record + ?Sized
```

### DIFF
--- a/noodles-bam/CHANGELOG.md
+++ b/noodles-bam/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+  * bam/record/codec: `decode` and `encode` functions are now public.
+
+  * bam/record/codec: Re-export `DecodeError` from the decoder module.
+
 ## 0.85.0 - 2025-12-11
 
 ### Changed
--- a/noodles-bam/src/record/codec.rs
+++ b/noodles-bam/src/record/codec.rs
@@ -1,6 +1,7 @@
-#![doc(hidden)]
+//! BAM record codec for encoding and decoding.
 
 pub mod decoder;
 pub mod encoder;
 
-pub(crate) use self::{decoder::decode, encoder::encode};
+pub use self::decoder::{DecodeError, decode};
+pub use self::encoder::encode;
--- a/noodles-bam/src/record/codec/decoder.rs
+++ b/noodles-bam/src/record/codec/decoder.rs
@@ -1,8 +1,10 @@
 //! BAM record decoder.
 
+#![allow(missing_docs)]
+
 mod bin;
 pub(crate) mod cigar;
-pub mod data;
+pub(crate) mod data;
 mod flags;
 mod mapping_quality;
 mod name;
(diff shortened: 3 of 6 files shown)


zaeleus:
> Using Writer::write_record() couples serialize+compress+write into one operation, preventing these stages from being parallelized independently.

Perhaps this is the misunderstanding? You can use the I/O interfaces to read from/write to any byte stream, not just a bgzip-compressed one, e.g., if you were working with raw, uncompressed records:

```rust
fn decode(src: &[u8], header: &sam::Header, record: &mut RecordBuf) -> io::Result<usize> {
	bam::io::Reader::from(src).read_record_buf(header, record)
}

fn encode<R>(dst: &mut Vec<u8>, header: &sam::Header, record: &R) -> io::Result<()>
where
	R: sam::alignment::Record,
{
	bam::io::Writer::from(dst).write_alignment_record(header, record)
}
```

nh13:
Thanks for the suggestion! I tried this approach, but the key issue is the decode side for my parallel pipeline use case.

My pipeline decompresses BGZF blocks in parallel, then a single thread parses records from the decompressed stream. Records can span block boundaries, so I need to:

1. Read the 4-byte block_size prefix manually
2. Buffer incomplete records across blocks
3. Decode just the record data (without prefix)

The codec's `decode(&mut &[u8], &mut RecordBuf)` is perfect for this: it advances the slice pointer, letting me track position through the byte stream.

With `Reader::from(src).read_record_buf()`, I'd need to ensure complete records are buffered before calling, which changes the architecture. The `&mut &[u8]` pattern that advances through the stream is more natural for cross-block record handling.

The encode side would work fine with `Writer::from(Vec::new())`.

Would you consider exposing just `decode()` with its current signature? Or is there a way to get the "advance slice pointer" behavior through the public API?

nh13:
@zaeleus I'm still interested in this and other PRs, let me know how I can help.  I know you're time is valuable, so let me know what I can do to make review the most efficient.

PR #365, opened 2026-01-02
fasta/io/indexed_reader: Use raw byte reading for faster queries

nh13:
## Summary

This changes `IndexedReader::query` to use raw byte reading with mathematical newline calculation instead of line-by-line parsing.

- Reads all bytes in one syscall, then strips newlines in memory
- Includes a fast path that skips newline stripping when reading within a single line
- ~10-20x improvement observed for multi-GB FASTA files

## Changes

- Added `indexed_reader/sequence.rs` with the core `read_sequence()` function
- Modified `IndexedReader::query()` to use the new implementation
- Changed trait bound from `BufRead + Seek` to `Read + Seek`

## Test plan

- [x] All existing tests pass
- [x] Added 8 new unit tests covering: full sequence, partial reads, line-spanning reads, single base, CRLF, empty, exact line boundaries, last partial line

### DIFF
--- a/noodles-fasta/Cargo.toml
+++ b/noodles-fasta/Cargo.toml
@@ -23,7 +23,12 @@ noodles-core = { path = "../noodles-core", version = "0.18.0" }
 tokio = { workspace = true, optional = true, features = ["fs", "io-util"] }
 
 [dev-dependencies]
+criterion = "0.5"
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+
+[[bench]]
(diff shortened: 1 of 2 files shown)


zaeleus:
> Reads all bytes in one syscall

`Read::read_exact` does not guarantee a single system call. I don't expect this to have have a significant effect, given the current implementation is buffered.

> ~10-20x improvement observed for multi-GB FASTA files

I'm unable to reproduce this. How are you comparing the implementations?

PR #371, opened 2026-01-24
feat(bgzf): add position tracking to MultithreadedWriter

nh13:
## Summary

Adds position tracking to `MultithreadedWriter` to enable building BAM indexes during multi-threaded compression.

## Motivation

When using `MultithreadedWriter` for parallel BAM compression, there's currently no way to determine the compressed file positions needed for BAI/CSI index construction. The standard `Writer` allows position tracking through its synchronous API, but `MultithreadedWriter` compresses and writes blocks asynchronously, making position correlation difficult.

This change enables building indexes during multi-threaded writes by:
1. Assigning sequential block numbers when blocks are sent for compression
2. Sending notifications (via channel) when blocks complete with their final compressed positions
3. Allowing callers to cache index entries with block numbers, then resolve positions when notifications arrive

## Use Case

This supports the parallel BAM processing pipeline I'm building (related to #364). The workflow:

1. Process records, caching index entries with `(block_number, uncompressed_offset)`
2. Flush blocks to compression (receive block number)
3. Receive `BlockInfo` notifications when blocks are written
4. Resolve cached index entries to final `(compressed_position, uncompressed_offset)` virtual offsets

## New Public API

- `BlockInfo` - struct with `block_number`, `compressed_start`, `compressed_size`, `uncompressed_size`
- `BlockInfoRx` - type alias for the receiver channel
- `block_info_receiver()` - get the notification receiver
- `current_block_number()` - next block number to be assigned
- `blocks_written()` - count of blocks fully written
- `position()` - current compressed file position
- `buffer_offset()` - bytes in staging buffer since last flush

## Alternatives Considered

- **Post-hoc index building**: Requires re-reading the BAM file after writing, doubling I/O for large files
- **Single-threaded Writer**: Works but sacrifices the compression parallelism that `MultithreadedWriter` provides

## Test Plan

- [x] Added unit test covering position tracking through multiple blocks
- [x] Existing tests pass

### DIFF
--- a/noodles-bgzf/src/io.rs
+++ b/noodles-bgzf/src/io.rs
@@ -13,8 +13,8 @@ pub mod writer;
 pub(crate) use self::block::Block;
 pub use self::{
     buf_read::BufRead, indexed_reader::IndexedReader, multithreaded_reader::MultithreadedReader,
-    multithreaded_writer::MultithreadedWriter, read::Read, reader::Reader, seek::Seek,
-    writer::Writer,
+    multithreaded_writer::{BlockInfo, BlockInfoRx, MultithreadedWriter},
+    read::Read, reader::Reader, seek::Seek, writer::Writer,
 };
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


holtjma:
Looks like there are some challenges here on functionality, but I will say +1 for having a way to using multithreaded writers with simultaneous index building.

PR #376, opened 2026-02-18
Avoid memory explosion in CRAM reading with Arc and multi-ref cache

txje:
The current implementation keeps reference sequences in a native Vec<u8> and clones the whole thing for every read in a slice. This is prohibitively slow and memory intensive (about 2TB RAM for a standard 10,000 read slice on human chromosome 1). We can avoid this by wrapping the fetched reference sequence in an Arc. For multi-ref slices, I also include a hashmap cache to avoid reloading the reference(s) for every read. This approach can tractably handle CRAM files of a realistic size.

### DIFF
--- a/noodles-cram/src/io/reader/container/slice.rs
+++ b/noodles-cram/src/io/reader/container/slice.rs
@@ -5,7 +5,7 @@ use std::{borrow::Cow, io};
 
 use noodles_core::Position;
 use noodles_fasta as fasta;
-use noodles_sam::{self as sam};
+use noodles_sam::{self as sam, alignment::Record as _};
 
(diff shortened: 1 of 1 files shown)


zaeleus:
Ah, thanks for the report.

Sequences used to be internally reference counted using [`bytes::Bytes`](https://docs.rs/bytes/1.11.1/bytes/struct.Bytes.html) but was changed in 3376caa185fe0151d67e27d395bfa02b386dcbe6. I even note to use a shareable type in this scenario but never applied it! The repository now returns a reference counted `Sequence` to avoid repeated cloning of sequences.

PR #387, opened 2026-04-10
Implement nibble-based decoding of BAM sequence

tfenne:
I've been looking for opportunities to shave time off the path of reading a BAM file into an iterator of RecordBuf.  This is a small focused change, and is internal to how the BAM 4-bit per base encoded sequence is inflated back into a Vec<u8>, using a nibble lookup rather than per-base shift and lookup.

The PR is organized into two commits - the first adds `criterion` as a dev dependency, and then sets up a BAM record decoding benchmark using what I think are fairly realistic short reads with a mixture of Cigars, and a pretty common set of aux tags per record.  The second then implements the nibble-decoding.  It's setup this way to enable easy running of the bench pre/post the change.  Also, if your preference is not to add criterion and/or benches to the repo, it should be easy to cherry-pick just the functional change.

The improvement is small, but real in my hands.  Testing on an aarch64 mac with the included bench I see about a 3% improvement in the overall time for the complete decoding of the BAM record into a RecordBuf.  This obviously doesn't include any I/O or bgzf decompression time - so real world gains will likely be smaller.

@zaeleus I'm more than happy to take feedback and re-work this if you'd like to see it done differently.  Given the small size and limited scope of the functional change I figured a PR would be easier than trying to describe this in an issue first.

### DIFF
--- a/noodles-bam/benches/read_record.rs
+++ b/noodles-bam/benches/read_record.rs
@@ -1,0 +1,196 @@
+use std::hint::black_box;
+use std::num::NonZero;
+
+use criterion::{Criterion, Throughput, criterion_group, criterion_main};
+use noodles_bam as bam;
+use noodles_core::Position;
+use noodles_sam::{
+    self as sam,
+    alignment::{
+        RecordBuf,
+        io::Write as _,
+        record::{
+            Flags, MappingQuality,
+            cigar::{Op, op::Kind},
+            data::field::Tag,
+        },
+        record_buf::{Cigar, Sequence, data::field::Value},
+    },
+    header::record::value::{Map, map::ReferenceSequence},
+};
+
+const NUM_RECORDS: usize = 10_000;
+
+fn build_cigar(i: usize, read_len: usize) -> Cigar {
+    let bucket = i % 100;
+
+    if bucket < 85 {
+        // 85%: perfect match
+        Cigar::from(vec![Op::new(Kind::Match, read_len)])
+    } else if bucket < 90 {
+        // 5%: 5' soft clip
+        Cigar::from(vec![
+            Op::new(Kind::SoftClip, 5),
+            Op::new(Kind::Match, read_len - 5),
+        ])
+    } else if bucket < 95 {
+        // 5%: 3' soft clip
+        Cigar::from(vec![
+            Op::new(Kind::Match, read_len - 5),
+            Op::new(Kind::SoftClip, 5),
+        ])
+    } else if bucket < 98 {
+        // 3%: single insertion
+        let half = read_len / 2;
+        Cigar::from(vec![
+            Op::new(Kind::Match, half),
+            Op::new(Kind::Insertion, 1),
+            Op::new(Kind::Match, read_len - half - 1),
+        ])
+    } else {
+        // 2%: single deletion
+        let half = read_len / 2;
+        Cigar::from(vec![
+            Op::new(Kind::Match, half),
+            Op::new(Kind::Deletion, 1),
+            Op::new(Kind::Match, read_len - half),
+        ])
+    }
+}
+
+fn build_md(i: usize, read_len: usize) -> Value {
+    let bucket = i % 100;
+
+    if bucket < 85 {
+        Value::from(format!("{read_len}"))
+    } else if bucket < 95 {
+        // soft clips don't appear in MD
+        let matched = read_len - 5;
+        Value::from(format!("{matched}"))
+    } else if bucket < 98 {
+        // insertion: MD reflects only ref-consuming ops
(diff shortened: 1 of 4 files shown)


noodles-bam/benches/read_record.rs
```diff
@@ -0,0 +1,196 @@
...
+                let mut n = 0u64;
+                while reader.read_record_buf(&header, &mut record).unwrap() > 0 {
+                    black_box(&record);
+                    n += 1;
+                }
+                n
+            })
+        });
+
+        group.bench_function("record_lazy", |b| {
```
zaeleus, line 177: Why is this described as lazy?

zaeleus:
Thanks, @tfenne. Your conclusion may be an example of hitting [Amdahl's law in the serial case](https://en.wikipedia.org/wiki/Amdahl%27s_law#Serial_programs), given the rather small speedup in your more integrated benchmark. When isolating the functions, there is a measureable improvement, which is certainly useful as the read length gets larger.

<details><summary><code>bench.rs</code></summary>

```rust
use std::hint::black_box;

use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};

fn decode_into_a(src: &[u8], base_count: usize, dst: &mut Vec<u8>) {
    let bases = src
        .iter()
        .flat_map(|&b| [decode_base(b >> 4), decode_base(b)]);

    dst.clear();
    dst.extend(bases);
    dst.truncate(base_count);
}

fn decode_base(n: u8) -> u8 {
    match n & 0x0f {
        0 => b'=', 1 => b'A', 2 => b'C', 3 => b'M',
        4 => b'G', 5 => b'R', 6 => b'S', 7 => b'V',
        8 => b'T', 9 => b'W', 10 => b'Y', 11 => b'H',
        12 => b'K', 13 => b'D', 14 => b'B', 15 => b'N',
        _ => unreachable!(),
    }
}

const NIBBLE_PAIRS: [[u8; 2]; 256] = {
    const BASES: [u8; 16] = *b"=ACMGRSVTWYHKDBN";

    let mut table = [[0u8; 2]; 256];
    let mut i = 0;

    while i < 256 {
        table[i] = [BASES[i >> 4], BASES[i & 0xf]];
        i += 1;
    }

    table
};

fn decode_into_b(src: &[u8], base_count: usize, dst: &mut Vec<u8>) {
    let bases = src.iter().flat_map(|&b| NIBBLE_PAIRS[b as usize]);

    dst.clear();
    dst.extend(bases);
    dst.truncate(base_count);
}

fn criterion_benchmark(c: &mut Criterion) {
    let mut dst = Vec::new();

    let mut group = c.benchmark_group("decode_into");

    for i in [150usize, 300] {
        group.bench_with_input(BenchmarkId::new("before", i), &i, |b, i| {
            let src = [0x12].repeat(i / 2);
            b.iter(|| decode_into_a(black_box(&src), black_box(*i), black_box(&mut dst)));
        });

        group.bench_with_input(BenchmarkId::new("after", i), &i, |b, i| {
            let src = [0x12].repeat(i / 2);
            b.iter(|| decode_into_b(black_box(&src), black_box(*i), black_box(&mut dst)));
        });
    }
}
```

</details> 

Abridged results (N = read length):

```
decode_into/before/150  time:   [24.130 ns 24.147 ns 24.167 ns]
decode_into/after/150   time:   [15.319 ns 15.358 ns 15.401 ns]
decode_into/before/300  time:   [45.070 ns 45.181 ns 45.339 ns]
decode_into/after/300   time:   [27.714 ns 27.799 ns 27.893 ns]
decode_into/before/1000 time:   [148.06 ns 148.34 ns 148.61 ns]
decode_into/after/1000  time:   [94.541 ns 94.913 ns 95.324 ns]
decode_into/before/10000
                        time:   [1.3906 µs 1.3918 µs 1.3932 µs]
decode_into/after/10000 time:   [868.02 ns 871.14 ns 874.71 ns]
```

The use of the lookup table is now merged in 779cadd568f763e13511adcc220b181e981a0f3c.

For further reference, your implementation is effectively the same as the current default/fallback in htslib: https://github.com/samtools/htslib/blob/1.23.1/sam_internal.h#L73-L89.

noodles-bam/benches/read_record.rs
tfenne: Oh, the second bench is here really just as a sanity check - that a second unaffected code path doesn't see any significant variance during A/B testing.  I called it "lazy" since this bench uses the bam::io::Reader directly which returns the bam::Record which doesn't parse most of the record (so doesn't hit the nibble decoding) but lazily decodes fields on demand (at least that's my understanding).

tfenne:
Thanks @zaeleus!

PR #389, opened 2026-04-24
bgzf: Use libdeflater::Crc when libdeflate feature is enabled

tfenne:
The block data checksum was previously always computed with flate2::Crc, even when libdeflate was selected for (de)compression. Routing CRC32 through libdeflater::Crc under the same feature matches the codec and is measurably faster on both x86_64 and aarch64. Applies to both the reader frame verification and writer trailer generation paths.

For context I've been spending a good amount of time profiling code that uses noodles to read BAM files.  When profiling on an apple silicon mac it jumped out at me that the time spent in CRC computation on the BAM decompression path seemed really high - in one run on a 1KG wgs BAM I was seeing 92s total for decompression, of which 19s was CRC computation.  

I've benchmarked this change on both apple silicon (which I expect to be comparable to graviton and other aarch64 platforms) and x86 (AVX512 at AWS).  The improvement on aarch64 is dramatic - the libdeflate CRC is 6.6x faster in my hands!  So that 92s decompression shrinks to ~76s.  On x86 it is less dramatic, but the libdeflate CRC is still over 2x faster there, which is still a nice win.

I made the matching change on the compression pathway too for symmetry's sake, although obviously the CRC computation is a much smaller fraction of the total on that path.

### DIFF
--- a/noodles-bgzf/CHANGELOG.md
+++ b/noodles-bgzf/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Changed
+
+  * bgzf/deflate: Use libdeflate's CRC32 implementation when the `libdeflate`
+    feature is enabled.
+
+    This matches the (de)compressor backend selection and is measurably faster
+    on both x86_64 and aarch64.
+
 ## 0.46.0 - 2026-02-18
 
 ### Changed
--- a/noodles-bgzf/src/deflate.rs
+++ b/noodles-bgzf/src/deflate.rs
@@ -78,10 +78,18 @@ pub(crate) fn encode(src: &[u8], compression_level: i32, dst: &mut Vec<u8>) -> i
     }
 }
 
+#[cfg(not(feature = "libdeflate"))]
 pub(crate) fn crc32(src: &[u8]) -> u32 {
     const START: u32 = 0;
 
     zlib_rs::crc32::crc32(START, src)
+}
+
+#[cfg(feature = "libdeflate")]
+pub(crate) fn crc32(src: &[u8]) -> u32 {
+    let mut crc = libdeflater::Crc::new();
+    crc.update(src);
+    crc.sum()
 }
 
 #[cfg(test)]


zaeleus:
Nice observations.

noodles uses zlib-rs for CRC32 by default.

To match libdeflate on x86-64 (assuming `target-features=+vpclmulqdq,+avx512f`), you do have to manually enable AVX512 support via [the `avx512` feature](https://github.com/trifectatechfoundation/zlib-rs/blob/v0.6.3/zlib-rs/Cargo.toml#L29). On an AMD Ryzen 9 9950X3D2:

```
zlib-rs/65536           time:   [738.38 ns 738.70 ns 739.18 ns]
libdeflate/65536        time:   [739.18 ns 739.55 ns 740.10 ns]
```

I'll likely raise noodles' MSRV to 1.89.0 and enable this by default.

I see that both zlib-rs and libdeflate use `__crc32*` intrinsics on AArch64, but libdeflate interleaves instructions. The latter is indeed a much faster approach. On an Apple M4:

```
zlib-rs/65536           time:   [5.5660 µs 5.5803 µs 5.6014 µs]
libdeflate/65536        time:   [756.19 ns 758.01 ns 759.84 ns]
```

It's fine to use the CRC32 function in libdeflate when it's enabled. I did recently remove flate2 from noodles-bgzf in a0fece90928ec87fc9d95e305c87f485d9f9012e, and it uses zlib-rs directly in 6fd1885e9d339cada70b8c714bc9d2b737ed1ad8. Can you rebase and make a specialized `noodles_bgzf::deflate::crc32` function when the `libdeflate` feature is enabled, i.e., similar to how `decode`/`encode` are done?

tfenne:
Ah, that's a good catch that enabling avx512 on zlib-rs essentially matches performance on x86!  I re-wrote my commit based on the latest `master` as you suggested (and overwrite the prior commit since that doesn't really rebase on master).

zaeleus:
Thanks!

PR #392, opened 2026-04-26
fasta/io/indexed_reader/builder: Add set_buffer_capacity for small perf gain

tfenne:
The uncompressed path of `Builder::build_from_path` previously wrapped the opened file in `BufReader::new`, which uses the std 8 KiB default capacity. When loading whole contigs via `read_sequence`, this ends up making tens of thousands of `read` syscalls per chromosome on a multi-GB reference.

The savings from this change are, realistically, quite small.  In my hands on an M1 mac, it does shave about 10% off with a buffer in the 32-128kb range, but in absolute terms, once the FASTA is in page cache, that's like 0.1-0.2s, so I'll understand if you don't think that's worthwhile.

This change:

  * adds `Builder::set_buffer_capacity` for callers that want to tune the buffer size, and
  * raises the default to 64 KiB, a friendly midpoint that captures most of the win without regressing on references with many tiny decoy/alt contigs (where bigger buffers waste read-ahead).

The BGZF branch is untouched: BGZF reads block-sized chunks directly from the file, and an outer `BufReader` was confirmed (separate bench) to provide no measurable benefit there.

I'd also be happy to make the buffer_size an Option and default to None (with the default then being the old behavior), or any other setup you'd prefer.

### DIFF
--- a/noodles-fasta/src/io/indexed_reader/builder.rs
+++ b/noodles-fasta/src/io/indexed_reader/builder.rs
@@ -10,10 +10,27 @@ use noodles_bgzf as bgzf;
 use super::IndexedReader;
 use crate::fai;
 
+/// Default read buffer capacity (in bytes) for uncompressed inputs.
+///
+/// Chosen empirically: on a multi-GB reference loaded contig-by-contig,
+/// throughput is flat between ~32 KiB and ~256 KiB and ~10% better than
+/// the 8 KiB [`BufReader`] default. 64 KiB is a friendly midpoint that
+/// also matches a typical filesystem block-cache page granularity.
+const DEFAULT_BUFFER_CAPACITY: usize = 64 * 1024;
+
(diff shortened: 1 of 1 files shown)


zaeleus:
Composing an indexed reader manually is likely a better solution than leaking its internals.

```rust
const DEFAULT_BUF_SIZE: usize = 64 * 1024;

let inner = File::open(&src).map(|f| BufReader::with_capacity(DEFAULT_BUF_SIZE, f))?;
let index = fai::fs::read(src.with_added_extension("fai"))?;
let mut reader = fasta::io::IndexedReader::new(inner, index);
```

tfenne:
That's fair.  I can certainly do that; I was just hoping to avoid duplicating a bunch of the code in the Builder that detects whether the input is compressed or not, the detects the necessary index file(s).

PR #404, opened 2026-07-14
vcf/header/parser: Honor header definitions of reserved keys

jdidion:
## Problem

`noodles-vcf`'s header record parser rejects any `##INFO`/`##FORMAT` definition of a *reserved* key whose `Number` (or `Type`) differs from the version-reserved default, but only for files declaring `##fileformat=VCFv4.3` or later. The same file declared as `VCFv4.2` parses fine, and htslib/bcftools accept it regardless.

Minimal repro (fails on 4.3/4.4/4.5, OK on 4.2):

```
##fileformat=VCFv4.4
##contig=<ID=chr1,length=248956422>
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
##FORMAT=<ID=DP,Number=.,Type=Integer,Description="d">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	S1
chr1	53210776	.	G	A	50	PASS	.	GT:DP	1/1:512
```

`read_header()` returns an `io::Error` ("invalid record") because `DP` is reserved as `Number=1`. `AD` declared as `Number=.` (reserved `Number=R`) fails the same way. These headers are legal VCF: a file may loosen a reserved field's `Number`.

## Fix

The header parser's `validate_format_definition`/`validate_info_definition` compared the header's declared `(Number, Type)` against the version-reserved default and errored on any difference. This is both spec-incorrect and inconsistent with the rest of the crate: both the `record_buf` and lazy `record` value readers already prefer the header's declared `(Number, Type)` over the reserved definition when parsing values, so the header was authoritative everywhere except this one gatekeeping check. This change drops that validation so an explicitly declared reserved key is honored. Keys the header does not declare are unaffected and still fall back to the version-reserved definition.

## Tests

Added regression tests: `parse_value` on redefined reserved FORMAT (`DP`, `AD` with `Number=.`) and INFO (`DP` with `Number=.`) keys under VCFv4.4, plus an end-to-end `Reader` test reading the minimal repro VCF. `cargo test`, `cargo fmt --check`, and `cargo clippy --all-features -- --deny warnings` all pass.

### DIFF
--- a/noodles-vcf/src/header/parser/record/value.rs
+++ b/noodles-vcf/src/header/parser/record/value.rs
@@ -10,6 +10,7 @@ use crate::header::{
 };
 
 /// An error returned when a VCF header record value fails to parse.
+#[allow(clippy::enum_variant_names)]
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum ParseError {
     InvalidFileFormat(string::file_format::ParseError),
@@ -20,28 +21,6 @@ pub enum ParseError {
     InvalidContig(map::contig::ParseError),
     InvalidOtherString(key::Other, string::ParseError),
     InvalidOtherMap(key::Other, map::other::ParseError),
-    FormatDefinitionMismatch {
-        id: String,
-        actual: (
-            crate::header::record::value::map::format::Number,
-            crate::header::record::value::map::format::Type,
-        ),
-        expected: (
-            crate::header::record::value::map::format::Number,
-            crate::header::record::value::map::format::Type,
-        ),
-    },
-    InfoDefinitionMismatch {
-        id: String,
-        actual: (
-            crate::header::record::value::map::info::Number,
-            crate::header::record::value::map::info::Type,
-        ),
-        expected: (
-            crate::header::record::value::map::info::Number,
-            crate::header::record::value::map::info::Type,
-        ),
-    },
 }
 
 impl error::Error for ParseError {
@@ -55,7 +34,6 @@ impl error::Error for ParseError {
             Self::InvalidContig(e) => Some(e),
             Self::InvalidOtherString(_, e) => Some(e),
             Self::InvalidOtherMap(_, e) => Some(e),
-            _ => None,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


zaeleus:
> These headers are legal VCF: a file may loosen a reserved field's `Number`.

This is not true; the specification disallows redefining reserved type definitions. Your example has an explicit failure in [`failed_meta_format_011.vcf`](https://github.com/samtools/hts-specs/blob/da617203a9527537746e200abda2885bec3a822c/test/vcf/4.3/failed/failed_meta_format_011.vcf).

To workaround this, you will need to parse the header as VCF 4.2. noodles does allow you to override the file format behavior by creating a custom header parser:

<details><summary><code>vcf_read_invalid_header.rs</code></summary>

```rust
use std::io::BufRead;

use noodles_vcf::{
    self as vcf,
    header::{FileFormat, parser::FileFormatOption},
};

const VCF_4_2: FileFormat = FileFormat::new(4, 2);

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let src = br#"##fileformat=VCFv4.4
##contig=<ID=chr1,length=248956422>
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
##FORMAT=<ID=DP,Number=.,Type=Integer,Description="d">
#CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO    FORMAT  S1
chr1    53210776        .       G       A       50      PASS    .       GT:DP   1/1:512"#;

    let mut reader = vcf::io::Reader::new(&src[..]);

    let header_reader = reader.header_reader();

    let mut parser = vcf::header::Parser::builder()
        .set_file_format_option(FileFormatOption::FileFormat(VCF_4_2))
        .build();

    for result in header_reader.lines() {
        let line = result?;
        parser.parse_partial(line.as_bytes())?;
    }

    let header = parser.finish();

    dbg!(&header);

    Ok(())
}
```

</details>

jdidion:
Thanks — you're right, and `failed_meta_format_011.vcf` confirms it: redefining a reserved key's Number is a spec conformance failure, so rejecting these records is correct behavior. The VCFs I hit (a clinical caller emitting `##FORMAT=<ID=DP,Number=.>` / `AD,Number=.`) are non-conformant; htslib is just being lenient. Forcing the parse to 4.2 via the header-parser `FileFormatOption` is exactly the right consumer-side handling — I'll use that. Closing this; the fix belongs on the producer. Appreciate the pointer.

Issue #213, opened 2023-10-24
Nondeterministic TBI index creation


holtgrewe:
I believe the use of `HashMap` in the following place in `noodles-csi/src/index/reference_sequence.rs` create nonderministic behaviour when creating TBI indices. I observe this when creating indices of the same `.vcf.gz` file multiple times and comparing the resulting binary index files.

```
/// A CSI reference sequence.
#[derive(Clone, Debug, Eq, PartialEq)]
pub struct ReferenceSequence {
    bins: HashMap<usize, Bin>,
    linear_index: Vec<bgzf::VirtualPosition>,
    metadata: Option<Metadata>,
}
```

**Edit** -- probably, `indexmap::IndexMap` would be a better choice here?

zaeleus:
Agreed, this can be changed to an ordered map to preserve insertion order, which will allow indices to be (re)serialized in the same way. Thanks for the suggestion!

PR #406, opened 2026-07-23
Compact binning indices (BAI/CSI/TBI) to match htslib/samtools

tfenne:
## Summary

`csi::binning_index::Indexer` builds spec-valid indices, but omits the two compaction behaviors htslib applies in `hts_idx_finish`. As a result it emits a larger index than `samtools index` for the same data — measured at ~1.3× on a deep-coverage WGS BAM and ~3.3× on a low-coverage (2×) WGS BAM — and carries more chunks per bin. This PR ports both behaviors so that every builder going through `Indexer` (BAI, CSI, TBI, and BCF's CSI) produces an index the same size and structure as samtools'. Query results are unchanged.

Both changes live in the shared `Indexer`, so all four formats benefit with no per-format code and no public API change.

## Change 1 — compact bins when building (`compress_binning`)

After the index is built, `ReferenceSequence::compact` folds any bin whose records span less than 64 KiB of compressed data into its parent bin, and merges chunks that share a BGZF block — mirroring htslib's `compress_binning` (`hts.c`). This drops the redundant fine-grained bins that the linear index already covers, which dominate the size for sparse data, and coalesces the chunk list for dense data.

The fold walks bins deepest-first (descending bin id) rather than level by level. Because `parent_id(b) = (b - 1) >> 3 < b`, descending order visits every bin after all of its descendants and before its parent, so a bin's span is final when tested and a parent is never removed before its children are considered — the same result as htslib's finest-to-coarsest cascade, but without needing the depth or the finest-level bin id, and touching each bin once. The linear index and per-bin linear offsets (CSI `loff`) are left untouched; folding is `loff`-safe because a parent's offset is already at or below its children's, and the CSI writer derives each surviving bin's `loff` from itself and its ancestors only.

## Change 2 — batch chunks of consecutive same-bin records (`hts_idx_push`)

`Indexer::add_record` now records a run of consecutive records assigned to the same bin as a single chunk spanning any BGZF block boundaries within the run, rather than one chunk per block — mirroring htslib's `hts_idx_push`. It anchors each record's chunk at the run's start so `Bin::add_chunk` coalesces the whole run; the linear index and metadata still see the record's own chunk.

Note on scope: the bundled `bam::fs::index` already avoids per-block splitting, because it advances each record's chunk start to the previous record's end (`start = end`), so a same-bin run coalesces across blocks automatically. So this change is effectively a no-op for the in-tree BAM indexer. Its value is (a) matching htslib's `hts_idx_push` batching semantics, and (b) making the `Indexer` robust for callers that compute per-record virtual positions independently: a record ending exactly at a BGZF block boundary otherwise starts a fresh chunk at every block. (A downstream index builder that computes chunk positions from block offsets saw ~18× index inflation on record-aligned BAMs from exactly this; this change eliminates that class of problem.) If you'd rather keep the `Indexer` minimal, Change 1 stands on its own — but I think Change 2 is worth it for htslib parity and downstream robustness.

## Results

Indexed with this branch vs `samtools index` (samtools 1.23.1) over the same file. "before" is the current `Indexer`; "this PR" is with both changes.

`HG03953.2x.bam` (4.15 GB, ~2× WGS subsampled from a deeper coverage 1000 genomes BAM):

| Metric | before | this PR | samtools |
| --- | ---: | ---: | ---: |
| BAI size | 9,020,384 B | 2,727,608 B | 2,727,608 B |
| Bins | 211,274 | 31,133 | 31,133 |
| Chunks | 351,867 | 48,639 | 48,639 |
| Linear intervals | 196,952 | 196,952 | 196,952 |
| n_no_coor | 8,282 | 8,282 | 8,282 |

`NA12813.bam` (67 GB, ~50× WGS, ~1.287B mapped reads):

| Metric | before | this PR | samtools |
| --- | ---: | ---: | ---: |
| BAI size | 12,313,472 B | 9,398,616 B | 9,398,616 B |
| Bins | 212,342 | 209,729 | 209,729 |
| Chunks | 555,608 | 374,736 | 374,736 |
| Linear intervals | 197,713 | 197,713 | 197,713 |
| n_no_coor | 352,968 | 352,968 | 352,968 |

In both cases the index is now the same size as samtools' and matches on every structural count. The reduction is all from Change 1: at low coverage many fine bins fall under the 64 KiB fold threshold and collapse into their parents (the dominant effect there); at high coverage fine bins hold enough data to be kept (as htslib intends) and the reduction comes from merging chunks that share a block.

The output is not byte-identical to samtools': the only difference is bin ordering within each reference (this PR keeps insertion order per #213; samtools writes hash order). Content, `idxstats`, and all queries are identical.

## Correctness & testing

Verified with each index swapped in under samtools, on both BAMs:

- `samtools idxstats` output identical (NA12813: 1,285,261,016 mapped / 1,881,446 unmapped).
- Region-query counts (`samtools view -c`) identical across a spread of regions (tiny windows, whole chromosomes, `chrM`, contig ends).
- Decoded records (`samtools view | md5`) byte-identical for the sampled regions.

New tests in the crate:

- Unit tests for the chunk helpers (`sort_chunks`, `compressed_span`, `merge_chunks_in_same_block`) and for `compact` (fold into an existing parent, keep a bin without a parent, keep a bin whose span is ≥ 64 KiB, never fold the root, and a cascade).
- Run batching: a run of same-bin records with chunks in different BGZF blocks collapses to one chunk; an unplaced record ends the run.
- A CSI (`BinnedIndex`) test builds through the `Indexer` so folding happens, then queries the folded-away leaf's region to confirm the linear-offset path stays correct.
- A test asserting the linear index records each record's own start (not the run's start) after batching.

Full workspace `cargo fmt --all -- --check`, `cargo clippy --all-features -- --deny warnings`, and `cargo test --all-features` pass, including the bam/tabix/bcf indexers that build through this `Indexer`. Changelog updated in `noodles-csi`.

## Notes

- Bin serialization stays in insertion order for determinism (#213); I did not sort bin ids on emit.
- As ever @zaeleus, I'd be happy to receive feedback and adjust this PR, and won't take offense if you decide to take the ideas from this PR and re-implement

### DIFF
--- a/noodles-csi/CHANGELOG.md
+++ b/noodles-csi/CHANGELOG.md
@@ -15,6 +15,14 @@
     reduces the size of the index for dense data (e.g. a human WGS BAI shrinks
     by roughly an order of magnitude). Query results are unchanged.
 
+  * csi/binning_index/indexer: Merge chunks of consecutive records in the same
+    bin.
+
+    A run of consecutive records assigned to the same bin is now stored as a
+    single chunk spanning any BGZF block boundaries within the run, rather than
+    one chunk per block, matching htslib's `hts_idx_push`. This reduces the
+    number of chunks per bin. Query results are unchanged.
+
 ## 0.57.0 - 2026-07-10
 
 ### Changed
--- a/noodles-csi/CHANGELOG.md
+++ b/noodles-csi/CHANGELOG.md
@@ -6,6 +6,15 @@
 
   * csi: Update to bit-vec 0.10.1.
 
+  * csi/binning_index/index/reference_sequence: Compact bins when building an
+    index.
+
+    After the index is built, a bin whose records span less than 64 KiB of
+    compressed data is folded into its parent bin, and chunks sharing a BGZF
+    block are merged, matching htslib's `compress_binning`. This substantially
+    reduces the size of the index for dense data (e.g. a human WGS BAI shrinks
+    by roughly an order of magnitude). Query results are unchanged.
+
 ## 0.57.0 - 2026-07-10
 
 ### Changed
--- a/noodles-csi/src/binning_index/index/reference_sequence.rs
+++ b/noodles-csi/src/binning_index/index/reference_sequence.rs
@@ -209,6 +209,16 @@ where
         self.index.last_first_start_position()
     }
 
+    // Adds or merges a chunk into the given bin, creating the bin if necessary.
+    pub(crate) fn add_chunk(&mut self, bin_id: usize, chunk: Chunk) {
+        self.bins
+            .entry(bin_id)
+            .or_insert(Bin::new(Vec::new()))
+            .add_chunk(chunk);
+    }
+
+    // Updates the linear index and metadata for a record. The chunk is the record's own, even when
+    // the bin stores a wider run chunk (see `Indexer::add_record`).
     pub(crate) fn update(
         &mut self,
         min_shift: u8,
@@ -218,10 +228,6 @@ where
         is_mapped: bool,
         chunk: Chunk,
     ) {
-        let id = reg2bin(start, end, min_shift, depth);
-        let bins = self.bins.entry(id).or_insert(Bin::new(Vec::new()));
-        bins.add_chunk(chunk);
-
         self.index.update(min_shift, depth, start, end, chunk);
 
         let metadata = self.metadata.get_or_insert(Metadata::new(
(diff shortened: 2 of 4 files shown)


zaeleus:
Interesting, thanks for mentioning it; I'll investigate further.

zaeleus:
This is now implemented in ae3b5dc5610f49abff4c060bfa90c15074a5073e. We'll forgo change 2. Thanks.

Issue #247, opened 2024-03-31
sam/io/writer/record: Increase visibility of SAM record CIGAR string serializer


zaeleus:
> @zaeleus thanks for all your work on this crate. What are the chances we can do the same and add `noodles::sam::alignment::record_buf::Cigar` with signature
> 
> ```Rust
> pub fn new(src: Vec<u8>) -> Self { ... }
> ```
> 
> And also expose a function 
> 
> ```Rust
> pub fn as_bytes(&self) -> &[u8]  { ... }
> ```
> 
> To both `...::record_buf:::Cigar` and `...::record::Cigar`?
> 
> It would be useful generally, and specifically for someone like me who wants to extract the CIGAR string and update the `MC` tag in BAM/SAM records.

_Originally posted by @theJasonFan in https://github.com/zaeleus/noodles/issues/233#issuecomment-2027767346_

---

The request here is to provide a method to serialize alignment record CIGAR operations to SAM CIGAR strings. This would enable using the serialized form in, e.g., record data string values.

@theJasonFan, the alignment record CIGAR buffer (`sam::alignment::record_buf::Cigar`) does not use a byte buffer for its internal representation, so your proposed methods won't quite work here. I think the better solution is to increase the visibility of the SAM record CIGAR field serializer, which will allow any implementation of `alignment::record::Cigar` to be serialized to a SAM CIGAR string.

zaeleus:
This is now available in noodles 0.68.0 / noodles-sam 0.56.0, e.g.,

```rust
// cargo add noodles@0.68.0 --features sam

use std::io;

use noodles::sam::{
    alignment::{
        record::{
            cigar::{op::Kind, Op},
            data::field::Tag,
        },
        record_buf::{data::field::Value, Cigar, Data},
    },
    io::writer::record::write_cigar,
};

fn main() -> io::Result<()> {
    let mut buf = Vec::new();
    let cigar: Cigar = [Op::new(Kind::Match, 4)].into_iter().collect();
    write_cigar(&mut buf, &cigar)?;

    let mut data = Data::default();
    data.insert(Tag::MATE_CIGAR, Value::String(buf.into()));

    eprintln!("{data:?}");
    // => Data([(Tag("MC"), String("4M"))])

    Ok(())
}
```

theJasonFan:
That was quick. Amazing! Thanks!

PR #411, opened 2026-08-05
cram/io: Increase the visibility of the variable-length integer codecs

BenjaminDEMAILLE:
`cram::io::reader::num` and `cram::io::writer::num` are `pub(crate)`, but the readers and writers they export (`read_itf8`, `read_itf8_as`, `read_ltf8`, `read_ltf8_as`, `read_uint7`, `read_uint7_as`, `write_itf8`, `write_ltf8`, `write_uint7`) are already `pub` inside those modules. The module visibility is what keeps them out of the public API.

ITF-8 and LTF-8 are named and defined by the CRAM specification independently of the container format, so they are part of the format rather than an implementation detail of this crate. Tooling built alongside noodles (container validators, index inspectors, format converters) currently has to reimplement them byte for byte.

This changes the two module declarations to `pub` and adds the documentation the public API requires (`missing_docs` is denied in CI), including an example per function. No signatures change.

This mirrors `sam::io::writer::record::write_cigar`, made public in #247 for the same reason.

Note on overlap: #380 also modifies `io/reader/num.rs` and `io/writer/num.rs`. This change is independent of that work: it only flips the two module declarations and documents what those modules already export, so it applies cleanly before or after it. It targets CRAM 3.0/3.1 as currently supported and makes no claim about any other version.

Split into three commits; each builds and passes `cargo fmt -- --check`, `cargo clippy --all-features -- --deny warnings`, and `cargo test --all-features` on its own.

### DIFF
--- a/noodles-cram/CHANGELOG.md
+++ b/noodles-cram/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## Unreleased
 
+### Added
+
+  * cram/io/reader: Add variable-length integer readers (`num::read_itf8`,
+    `num::read_itf8_as`, `num::read_ltf8`, `num::read_ltf8_as`,
+    `num::read_uint7`, and `num::read_uint7_as`) ([#411]).
+
+  * cram/io/writer: Add variable-length integer writers (`num::write_itf8`,
+    `num::write_ltf8`, and `num::write_uint7`) ([#411]).
+
+[#411]: https://github.com/zaeleus/noodles/pull/411
+
 ### Changed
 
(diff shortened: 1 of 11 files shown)


BenjaminDEMAILLE:
@zaeleus one practical note: the CI checks are not running on this because first-time contributors need workflow approval. They pass locally, per commit rather than only on the tip: `cargo fmt -- --check`, `cargo clippy --all-features -- --deny warnings`, and `cargo test --all-features` (234 unit, 139 doc).

No rush on any of this, and no objection if you would rather take the one-line visibility change and write the docs yourself.

PR #412, opened 2026-08-05
cram/codecs: Increase the visibility of the rANS decoders and encoders

BenjaminDEMAILLE:
`codecs::rans_4x8` and `codecs::rans_nx16` are public modules, but their `decode` and `encode` functions are `pub(crate)`. What the modules currently export is `Order` and `Flags`: the parameter types for operations that cannot be called.

rANS 4x8 and rANS Nx16 are specified by CRAMcodecs independently of the CRAM container format, and htscodecs ships them as a standalone library used outside CRAM. There is no pure-Rust implementation of either that is usable as a library today.

This changes the two re-exports to `pub` and adds the documentation the public API requires (`missing_docs` is denied in CI), with a round-trip example per function. No signatures change, and the `decode`/`encode` modules themselves stay crate-private, so the internal helpers next to them (`state_step`, `bit_pack`, and so on) are not exposed.

Split into three commits; each builds and passes `cargo fmt -- --check`, `cargo clippy --all-features -- --deny warnings`, and `cargo test --all-features` on its own.

Kept separate from #411 on purpose: two modules, two reasons, and a refusal on one should not carry the other.

### DIFF
--- a/noodles-cram/CHANGELOG.md
+++ b/noodles-cram/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Added
+
+  * cram/codecs/rans_4x8: Add decoder (`decode`) and encoder (`encode`)
+    ([#412]).
+
+  * cram/codecs/rans_nx16: Add decoder (`decode`) and encoder (`encode`)
+    ([#412]).
+
+[#412]: https://github.com/zaeleus/noodles/pull/412
+
 ### Changed
 
   * cram: Update to lzma-rust2 0.20.0.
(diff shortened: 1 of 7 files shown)


BenjaminDEMAILLE:
@zaeleus same note as on #411: CI is awaiting workflow approval for a first-time contributor. Locally, each of the three commits passes `cargo fmt -- --check`, `cargo clippy --all-features -- --deny warnings`, and `cargo test --all-features` on its own.

The two PRs are independent, so taking one and declining the other is fine.

PR #414, opened 2026-08-05
vcf: Add an option to write header records in the order they were added

### LINKED ISSUE #413: vcf: Reading and writing a header does not reproduce the input
`io::writer::header::write_header` writes records grouped by category: `INFO`, then `FILTER`, `FORMAT`, `ALT`, `contig`, and finally other records. `Header` keeps insertion order *within* a category, because each category is an `IndexMap`, but nothing relates the categories to each other.

Reading a VCF header and writing it back therefore does not reproduce the input, even when nothing was modified. The round trip is not idempotent.

```rust
#[test]
fn test_write_header_round_trip() -> io::Result<()> {
    const RAW_HEADER: &str = "##fileformat=VCFv4.5
##FILTER=<ID=q10,Description=\"Quality below 10\">
##INFO=<ID=NS,Number=1,Type=Integer,Description=\"Number of samples\">
#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO
";

    let header: Header = RAW_HEADER.parse().map_err(io::Error::other)?;

    let mut buf = Vec::new();
    write_header(&mut buf, &header)?;

    assert_eq!(buf, RAW_HEADER.as_bytes());

    Ok(())
}
```

The `FILTER` record comes back after the `INFO` record. Nothing in the VCF specification requires a particular order, so the output stays valid, but the file is not the one that was read.

Two shapes seem possible:

1. `Header` retains global insertion order and `write_header` follows it.
2. A writer option selects between grouped, the current behavior, and insertion order.

Both are observable behavior changes if insertion order becomes the default, so 2 with grouped as the default is the conservative one.

I have a branch implementing 2: `Header::record_order` returns the record keys in the order they were added, populated by the builder and the parser, and `io::writer::Builder::set_record_order` chooses between `RecordOrder::Grouped` (default, unchanged) and `RecordOrder::AsAdded`. Records added through the mutable accessors are not tracked, so they are written last, grouped by category, and every record is still written exactly once. One wrinkle worth flagging: `Header::PartialEq` has to ignore the new field, otherwise two headers with the same records compare unequal because they were built in a different order.

Happy to open it as a PR, adjust the shape, or drop it if you would rather this stay as it is.

Context: found while working on <https://github.com/IPNP-BIPN/htsjdk-rs>, a byte-identical Rust port of htsjdk, where the VCF write path has to reproduce its input.

BenjaminDEMAILLE:
Closes #413.

`io::writer::header::write_header` writes records grouped by category: `INFO`, then `FILTER`, `FORMAT`, `ALT`, `contig`, and finally other records. `Header` keeps insertion order *within* a category, because each category is an `IndexMap`, but nothing relates the categories to each other. Reading a VCF header and writing it back therefore does not reproduce the input, even when nothing was modified.

This adds `Header::record_order`, the record keys in the order they were added, populated by the builder and the parser, and `io::writer::Builder::set_record_order`, which selects between:

- `RecordOrder::Grouped`, the current behavior and the default, so nothing changes for existing users;
- `RecordOrder::AsAdded`, which follows `Header::record_order` and makes the round trip idempotent.

Records added through the mutable accessors, e.g., `Header::infos_mut`, are not tracked. They are written last, grouped by category, so a header mutated that way still writes every record exactly once and no key is lost. A key that no longer resolves to a record is skipped.

Two things worth flagging before you find them:

1. `Header::PartialEq` is now implemented manually and ignores the record order. Without that, two headers holding the same records compare unequal because they were built in a different order, which would break existing callers and several tests in this repository.
2. This is only a behavior change if `AsAdded` is selected. `Writer::new` and `Builder::default` both keep `Grouped`.

Split into three commits: the header change, the writer change, the changelog. Each builds and passes `cargo fmt -- --check`, `cargo clippy --all-features -- --deny warnings`, and `cargo test --all-features` across the whole workspace on its own.

Three tests cover it: the round trip under `AsAdded`, the unchanged grouped output, and the fallback for records added through a mutable accessor.

I opened #413 first rather than sending this straight over, and I am still happy to drop it, change its shape, or let you write it your own way.

Context, not justification: noticed while using noodles-vcf as an independent reader to cross-check a VCF writer in <https://github.com/IPNP-BIPN/htsjdk-rs>. The defect stands on its own regardless of that; I mention it only so you know where it came from.

### DIFF
--- a/noodles-vcf/CHANGELOG.md
+++ b/noodles-vcf/CHANGELOG.md
@@ -11,11 +11,28 @@
   * vcf/variant/record_buf/alternate_bases: Implement `Extend<String>` and
     `FromIterator<String>`.
 
+  * vcf/header: Add record order (`Header::record_order`).
+
+    This is the keys of the records in the order they were added. Records added
+    using the mutable accessors, e.g., `Header::infos_mut`, are not tracked.
+
+  * vcf/io/writer: Add header record order option
+    (`writer::Builder::set_record_order`).
+
+    `RecordOrder::Grouped` (default) groups records by category, which is the
+    previous behavior. `RecordOrder::AsAdded` follows `Header::record_order`,
+    which makes reading and writing a header idempotent.
+
 ### Changed
 
   * vcf/record/samples/series/value/genotype: Return error on empty input.
 
     Empty inputs now return `io::ErrorKind::UnexpectedEof`.
+
+  * vcf/header: Exclude the record order from `Header` equality.
+
+    `PartialEq` is now implemented manually. Two headers with the same records
+    remain equal regardless of the order they were added in.
 
 ## 0.93.0 - 2026-08-21
 
@@ -1277,7 +1294,6 @@
   * vcf/record/genotypes/genotype/field: Remove `Field`.
 
     Use `(Key, Option<Value>)` instead.
-
 
 ## 0.23.0 - 2022-11-29
 
--- a/noodles-vcf/src/header.rs
+++ b/noodles-vcf/src/header.rs
@@ -4,11 +4,12 @@ mod builder;
 pub mod file_format;
 pub mod parser;
 pub mod record;
+mod record_key;
 pub mod string_maps;
 
 pub use self::{
     builder::Builder, file_format::FileFormat, parser::ParseError, parser::Parser, record::Record,
-    string_maps::StringMaps,
+    record_key::RecordKey, string_maps::StringMaps,
 };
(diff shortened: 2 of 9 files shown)


BenjaminDEMAILLE:
@zaeleus this is the branch from #413, opened so the diff is visible rather than described. The question in the issue still stands: if you would rather leave the write order as it is, or design the option differently, closing this costs nothing.

PR #415, opened 2026-08-05
cram: Add access to the raw blocks of a container slice

BenjaminDEMAILLE:
`Slice::decode_blocks` is currently the only way to reach the blocks of a slice. It decodes every block eagerly and calls `read_block_as`, which rejects any block whose content type is not the one the specification expects. That is the right behavior for reading records, but it means a container whose blocks are wrong cannot be inspected at all: the thing you want to look at is the thing that makes the call fail.

`Slice::blocks` returns the blocks as they appear in the stream, with their compression method, content type, content ID, uncompressed size, and compressed source. Nothing is decoded, `Block::decode` does that on demand, and the block CRC32 is still verified while reading.

Getting there needs two visibility changes, both of the same kind as #411 and #412:

- `container::block` is `pub(crate)`, but `CompressionMethod` and `ContentType` inside it are already `pub`, fully documented, and marked `#[non_exhaustive]`. They read as items written to be public and then hidden by the module declaration. Both are defined by the CRAM specification in § 8 "Block structure".
- `Block` is re-exported as `io::reader::Block`, next to `Container`. Without it the type is returned by a public method but cannot be named, and `rustc` flags the re-export inside `io::reader::container` as unused because that module is itself `pub(crate)`.

While doing this I noticed a related gap I deliberately left alone: `Slice` is returned by `Container::slices` and has the same problem, reachable through method calls but impossible to name. Happy to fix that too, in a separate change, if you agree it is one.

`read_block` and `read_block_as` are now both `pub(crate)`; neither was reachable before, so no public signature changes.

Split into three commits: the container change, the reader change, the changelog. Each builds and passes `cargo fmt -- --check`, `cargo clippy --all-features -- --deny warnings`, and `cargo test --all-features` across the whole workspace on its own.

Note on overlap: the changelog entry adds an `## Unreleased` section, as do #411 and #412. Whichever lands first, I will rebase the others.

### DIFF
--- a/noodles-cram/CHANGELOG.md
+++ b/noodles-cram/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 ## Unreleased
 
+### Added
+
+  * cram/container: Add block module (`container::block`).
+
+    This exposes `block::CompressionMethod`, `block::ContentType`, and
+    `block::ContentId`, which were already written as public items.
+
+  * cram/io/reader/container/slice: Add raw blocks iterator
+    (`Slice::blocks`).
+
+    Unlike `Slice::decode_blocks`, this decodes nothing and does not require
+    the block content types to be what the specification expects, which makes
+    it usable to inspect or validate a container.
+
+  * cram/io/reader: Add container block (`io::reader::Block`).
+
 ### Changed
 
   * cram: Update to lzma-rust2 0.20.0.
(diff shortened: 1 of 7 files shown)


BenjaminDEMAILLE:
@zaeleus CI here is also awaiting workflow approval. Locally each of the three commits passes `cargo fmt -- --check`, `cargo clippy --all-features -- --deny warnings`, and `cargo test --all-features` on its own.

This is the one of the four I am least attached to, since it adds API rather than correcting an inconsistency. If you would rather shape the raw-block access differently, or not have it, say so and I will close it.

PR #418, opened 2026-08-05
bam: Make the BAI reader and writer usable without noodles-sam

### LINKED ISSUE #417: bam: Reading a BAI requires noodles-sam, which the index code does not use
Reading a `.bai` requires depending on `noodles-bam`, which brings `noodles-sam` and everything under it, although the index code needs none of it.

Measured on 0.115.0. `noodles-bam/src/bai` is 35 files, and none of them names `noodles_sam`:

```console
$ grep -rl noodles_sam noodles-bam/src/bai
$ # no output
```

Its couplings are `noodles-csi` for the binning index model, and, inside this crate, `crate::io::reader::num` and `crate::io::writer::num` for the little-endian primitives. That is all.

The cost of the difference, unique crates in the normal dependency graph:

| | crates |
|---|---|
| `noodles-csi` | 18 |
| `noodles-bam` | 27 |

So a consumer that only wants to parse an index compiles `noodles-sam` and nine further crates for it, including `lexical-core` and its six companions, which exist for SAM text parsing.

I have no strong view on the shape, and each option costs you something I cannot judge from outside:

1. Make `noodles-sam` optional behind a default-on feature that gates the record, reader, writer and `fs` modules, leaving `bai` reachable with `--no-default-features`. Cheapest, but it puts a large part of the crate behind `#[cfg]`, which is a maintenance tax on every future change.
2. Move the index to its own crate, next to `noodles-csi` and `noodles-tabix`, which are already separate for the same reason. Cleanest by the layout the workspace already uses, and a breaking move.
3. Decide it is only a build cost and leave it. Entirely reasonable, and worth writing down if so, because the asymmetry with `noodles-tabix` reads as an oversight rather than a decision.

I am happy to implement whichever you prefer, or none. Filing rather than sending a PR because guessing wrong here wastes your review rather than mine.

Related: #416 removes rayon and crossbeam-channel from the same graph when the multithreaded I/O is not wanted. The two are independent.

BenjaminDEMAILLE:
Closes #417, taking the first of the three options in it.

Two commits of substance, and the first stands on its own regardless of what you think of the second.

**`bam/bai`: decouple the index from the record reader and writer.** The index read its integers with `io::reader::num` and wrote them with `io::writer::num`, which are private helpers of the record reader and writer. That is the wrong direction: BAI is defined independently of records, and the index knows nothing about them. It now has its own `bai::io::num`. The proof that these were the index's helpers all along is that `read_u64_le` and `write_u64_le` become dead code once it stops calling them, so they are removed.

**`bam`: gate everything but the index behind a `records` feature.** Enabled by default, so nothing changes unless it is turned off. It gates `record`, `io`, and `fs`; `bai` is always available. `async` implies `records`, since the asynchronous modules read and write records.

The normal dependency graph, measured with `cargo tree -e normal`:

| | crates |
|---|---|
| default | 27 |
| `--no-default-features` | 19 |

The eight that leave are `noodles-sam`, `bitflags`, and the six `lexical-*` crates, which exist for SAM text parsing.

## What it costs you, which #417 said I could not judge from outside

Now that it is built, I can. The tax is not the four `#[cfg]` lines in `lib.rs`; it is the periphery:

- 16 `[[example]]` blocks gained a `required-features`, mirroring what the asynchronous examples already do. That is 48 lines of manifest.
- The two crate-level doctests needed an explicit `fn main` under `#[cfg]`, because a doc example cannot otherwise be conditional. The added lines are all `#`-prefixed, so nothing changes in the rendered documentation, but they are there in the source.
- Four small helpers now exist in two places. A shared crate-root `num` would avoid that, but moving it touches 54 call sites across 45 files of unrelated modules, which is a worse diff than the duplication.

If you would rather have option 2 from the issue, the index in its own crate next to `noodles-csi` and `noodles-tabix`, this PR is the wrong shape and I will close it without argument. If you would rather have option 3, leave it as it is, the first commit is still worth taking on its own and I will reduce the PR to that.

Verified in four configurations: `--no-default-features` (7 unit, 17 doc), default (133, 64), `--all-features` (154, 102), and the whole workspace at `--all-features`, plus `cargo fmt -- --check` and `cargo clippy --deny warnings` both with and without default features. All three commits pass all of it on their own.

### DIFF
--- a/noodles-bam/CHANGELOG.md
+++ b/noodles-bam/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+  * bam: Add `records` feature.
+
+    This gates the record (`record`), I/O (`io`), and filesystem operations
+    (`fs`) modules, i.e., everything but the index (`bai`). It is enabled by
+    default, so the previous behavior is unchanged. Disabling it removes the
+    noodles-sam dependency, which the index does not use.
+
+### Changed
+
+  * bam/bai: Read and write integers using the index's own helpers.
+
+    The index previously borrowed `io::reader::num` and `io::writer::num`,
+    which belong to the record reader and writer.
+
 ## 0.95.0 - 2026-08-21
 
 ### Changed
--- a/noodles-bam/Cargo.toml
+++ b/noodles-bam/Cargo.toml
@@ -12,13 +12,16 @@ documentation = "https://docs.rs/noodles-bam"
 categories = ["parser-implementations", "science::bioinformatics"]
 
 [features]
+default = ["records"]
 async = [
   "dep:futures",
   "dep:pin-project-lite",
   "dep:tokio",
   "noodles-bgzf/async",
   "noodles-csi/async",
+  "records",
 ]
+records = ["dep:noodles-sam"]
 
 [dependencies]
 bstr.workspace = true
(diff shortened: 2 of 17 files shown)


BenjaminDEMAILLE:
@zaeleus CI is awaiting workflow approval, as on the others. Locally all three commits pass `cargo fmt -- --check`, `cargo clippy --deny warnings` with and without default features, `cargo test --all-features` across the workspace, and `cargo test -p noodles-bam --no-default-features`.

Order of preference if you want only part of it: the first commit alone, then the whole thing, then nothing.

Issue #421, opened 2026-08-05
sam/record: Only one of the four record ref hooks is overridden


BenjaminDEMAILLE:
`bam::Record` overrides all four of the `#[doc(hidden)]` hooks on `alignment::record::Record`: `cigar_ref`, `sequence_ref`, `quality_scores_ref` and `data_ref`. `sam::Record` overrides one, `sequence_ref`. The other three fall through to the defaults, which box.

#419 covers `quality_scores_ref`, where the fix is exact because `QualityScoresRef::Offset` describes SAM quality scores precisely. This issue is about what is left, which is harder and which I have not sent a patch for.

## Where the time goes after #419

Same machine and input as #419: Apple M4 Max, 1,604,108-record SAM, release with `lto = "fat"`. Encode alone, by subtracting a parse-only run from a run writing to `io::sink()`, is 0.782 s. Sampling the encode loop, of 2,723 main-thread samples:

- **882 in `write_data`**, by far the largest item. Inside it: `parse_field` and `parse_string` re-parsing each tag from the text, drop glue for `Value`, and one boxed iterator per record from `Data::iter`.
- The rest is spread thinly across the remaining fields.

For comparison, htslib spends about 0.09 s on the equivalent work for a similar file, because `sam_parse1` parses the text straight into the BAM binary layout. noodles parses into a `sam::Record`, then walks it and encodes. The tags are visited twice.

## Why the existing variants do not help

`DataRef::FieldEncoded(&[u8])` and `CigarRef::FourBytePacked(&[u8])` both take an already-BAM-shaped slice, which is what `bam::Record` has and `sam::Record` never will. SAM stores tags and CIGAR as text, so this is genuine transcoding, not a missed borrow. There is no `Offset`-shaped trick available.

What might work, and what I am asking about rather than proposing:

1. A text-shaped variant, e.g. `DataRef::Text(&[u8])`, with the parse living in noodles-sam and the encode in noodles-bam, so the fields are visited once. This adds a variant to a `#[doc(hidden)]` enum and moves a parser boundary, which is a design call rather than a patch.
2. Leaving it. The two-pass structure is what makes `sam::Record` lazy and cheap to read, which is why noodles' parser is roughly three times faster than htslib's in isolation, 0.229 s against 0.45 s on comparable input. The current design wins on read and loses on convert, and that may well be the right trade.

[...]

Issue #420, opened 2026-08-05
bgzf: MultithreadedWriter's worker count plateau deserves a sentence in the docs


BenjaminDEMAILLE:
`bgzf::io::MultithreadedWriter` stops getting faster after a small number of workers, and the reason is not in the writer. I spent a while looking for a bounded queue serialising the producer before measuring the obvious thing, so this is a request for one sentence of documentation rather than a bug report.

Apple M4 Max, 12 performance and 4 efficiency cores, rustc 1.97.1, release with `lto = "fat"`, `libdeflate` enabled. Converting a 1,604,108-record SAM to BAM:

| workers | wall |
|---|---|
| 1 | 3.60 s |
| 2 | 1.89 s |
| 4 | 1.09 s |
| 8 | 1.08 s |
| 16 | 1.09 s |

The same loop writing to `io::sink()` instead of a BGZF writer, so parse and encode only, takes **1.011 s**. The plateau is 1.08 s. The writer is contributing about 70 ms and everything else is the single-threaded producer that feeds it.

That is the whole explanation. By four workers, compression is off the critical path, and no further worker can help because the caller is producing records serially. A user who asks for sixteen workers and measures no change will look for the fault in the writer, which is where I looked.

Two things that would have saved the search, either of which is enough:

1. A sentence on `MultithreadedWriter` and `with_worker_count` saying that workers only compress, and that throughput is bounded by how fast the caller can hand over uncompressed blocks.
2. The same sentence noting that past the point where compression keeps up, adding workers is free but idle.

Related: #419 lowers the producer cost for the SAM to BAM case, which moves the plateau from 1.32 s to 1.09 s without touching this crate. That is the shape of every fix here: the writer was never the wall.

BenjaminDEMAILLE:
@zaeleus this is the cheapest of the three I filed today: no code, one sentence. I went looking for a bounded queue or a channel-depth problem in the writer before measuring the producer, and the producer was the whole answer. The note would stop the next person doing the same.

PR #423, opened 2026-08-05
bam/record/codec/encoder: Read the CIGAR operations once per record

BenjaminDEMAILLE:
Encoding one record walks the CIGAR four times:

| line | call | why |
|---|---|---|
| `encoder.rs:101` | `Record::alignment_end` | the bin, via `alignment_span` |
| `encoder.rs:106` | `Cigar::len` | the operation count |
| `encoder.rs:139` | `write_cigar` | the operations themselves |
| `encoder.rs:143` | `Cigar::read_length` | validating the sequence length |

Each of the first, second and fourth goes through `Record::cigar`, which returns `Box<dyn Cigar>`, so it is four allocations as well as four walks. For a `sam::Record` a walk is a parse of the text, so a record with a CIGAR of `95M2I4M` parses it four times to write it once.

The span, the read length and the operation count all fall out of one pass, so this takes one. `Record::sequence_ref` replaces `Record::sequence` for the base count, which drops a fifth allocation, and the operations are still written through `Record::cigar_ref`, so `bam::Record`'s packed fast path is untouched.

`overflowing_write_cigar_op_count` now takes the counts instead of the CIGAR, since it no longer needs to walk it.

## Measurement

Apple M4 Max, 12 performance and 4 efficiency cores, 128 GB, rustc 1.97.1, release with `lto = "fat"` and `codegen-units = 1`, `noodles-bgzf/libdeflate`. Encode is isolated by subtracting a parse-only run from a run writing to `io::sink()`. Best of three, page cache warm.

| encode alone | before | after |
|---|---|---|
| 1,604,108 SAM records | 0.985 s | 0.891 s |
| 654,610 BAM records | 0.383 s | 0.273 s |

The BAM side gains more because its CIGAR walk was already cheap per operation, so the four allocations dominated.

Stacked with #419, which is independent, encoding the same SAM drops to 0.646 s and the same BAM to 0.266 s. BAM to BAM at 16 workers goes from 0.494 s to 0.419 s, against 0.45 s for `samtools view -b -@ 16` on the same file and machine.

## Correctness

Byte-identical output in both directions, `cmp` on the whole file: SAM to BAM and BAM to BAM against what noodles produced before the change. The full workspace test suite passes with `--all-features`.

Two commits: the encoder change, and the changelog.

### DIFF
--- a/noodles-bam/CHANGELOG.md
+++ b/noodles-bam/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Changed
+
+  * bam/record/codec/encoder: Read the CIGAR operations once per record.
+
+    The alignment span, the read length, and the operation count now come from
+    a single pass rather than one each, which also removes four allocations per
+    record. Output is unchanged.
+
 ## 0.95.0 - 2026-08-21
 
 ### Changed
--- a/noodles-bam/src/record/codec/encoder.rs
+++ b/noodles-bam/src/record/codec/encoder.rs
@@ -15,7 +15,8 @@ mod template_length;
 
 use std::{error, fmt, io};
 
-use noodles_sam::{self as sam, alignment::Record};
+use noodles_core::Position;
+use noodles_sam::{self as sam, alignment::Record, alignment::record::Cigar};
 
 use self::{
     bin::write_bin,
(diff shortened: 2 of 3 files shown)


BenjaminDEMAILLE:
@zaeleus this stacks with #419 and neither depends on the other. Together they take the encode of the same SAM from 0.985 s to 0.646 s, and BAM to BAM at 16 workers from 0.494 s to 0.419 s, against 0.45 s for `samtools view -b -@ 16` on the same file and machine.

What they do not do is change how the curve scales, which is #424, and that one is a question rather than a proposal because the obvious answer turned out to be wrong when I measured it.

BenjaminDEMAILLE:
Related issues: #421 is what remains in the encode path after this and #419, #424 is why the conversion stops scaling, #420 is the documentation sentence for that plateau. #427 is the caller-side pattern that does scale.

Issue #424, opened 2026-08-05
sam,bam: SAM to BAM conversion does not scale with cores, and parallelising the caller only buys 20%


BenjaminDEMAILLE:
Following on from #419, #423 and #420. Those cut the serial cost of converting SAM to BAM by about a third and explain the worker plateau, but they do not change the shape of the curve: throughput stops responding to cores because one thread parses and encodes every record while the BGZF pool only compresses.

`samtools view -b -@ N` keeps scaling on the same input and machine, because htslib parallelises the record pipeline itself, not only the deflate.

Apple M4 Max, 12 performance and 4 efficiency cores, rustc 1.97.1, release with `lto = "fat"`, `libdeflate`. 1,604,108-record SAM, 620 MB. Best of three.

| BGZF workers | noodles today | with #419 and #423 | `samtools view -b -@ N` |
|---|---|---|---|
| 1 | 3.63 s | 3.60 s | 5.20 s |
| 4 | 1.32 s | 0.99 s | 1.38 s |
| 8 | 1.33 s | 0.98 s | 0.71 s |
| 16 | 1.33 s | 0.97 s | 0.46 s |

noodles wins to four workers and loses past six. The crossover is where compression stops being the constraint.

## What I measured before proposing anything

I wrote a caller-side parallel producer to check whether the missing scaling can simply be done outside the library: read the text in 16 MB chunks, split on newlines, parse and encode line ranges on a rayon pool, write the encoded blocks in order. Output is byte-identical to the serial path.

| configuration | wall |
|---|---|
| serial producer, 16 BGZF workers | 0.97 s |
| parallel producer, 16 rayon + 16 BGZF | 0.81 s |
| parallel producer, 16 rayon + 32 BGZF | 0.77 s |
| parallel producer, 8 rayon + 8 BGZF | 0.98 s |

**Parallelising the producer buys 20%, not a factor of the core count.** That result is the reason I am filing an issue rather than a design. Something after the encode is absorbing the gain, and the candidates I can see from outside are the copy into the writer's block buffer and the ordered handoff. Whatever it is, "spawn threads around the encoder" is not the answer, and anyone who assumes it is, as I did, will spend an afternoon finding out.

So this is a question rather than a request:

1. Is a threaded record pipeline something you want in noodles at all? It is a large piece of surface, and "noodles is the fast single-threaded one, compose it yourself" is a legitimate answer that would be worth stating in the docs next to #420.
[...]

PR #426, opened 2026-08-05
sam/io/writer/record: Write the sequence and quality scores in blocks

BenjaminDEMAILLE:
Writing a SAM record emits `SEQ` and `QUAL` one byte at a time:

```rust
// io/writer/record/quality_scores.rs
for result in quality_scores.iter() {
    let n = result?;
    // ...
    writer.write_all(&[m])?;
}
```

and the same shape in `write_four_bit_packed_sequence` and `write_raw_sequence`. A 100 bp read is 200 `write_all` calls, each a call and a bounds check through the `BufWriter`, and for a 654,610-record BAM that is about 130 million of them.

Two changes:

- Both fields are filled into a 256-byte block and written a block at a time. A raw sequence, which is already the output bytes, is written straight from its slice.
- `write_quality_scores` takes a `QualityScoresRef` rather than a boxed `QualityScores`. `bam::Record` already returns `QualityScoresRef::Raw` and `sam::Record` returns `QualityScoresRef::Offset`, so both lose the boxed iterator and the per-base dynamic dispatch. Scores already stored with SAM's own offset are written without being touched, which is the SAM-to-SAM case.

`record.sequence()` also becomes `record.sequence_ref()` for the base count, which drops one allocation per record.

## Measurement

Apple M4 Max, 12 performance and 4 efficiency cores, rustc 1.97.1, release with `lto = "fat"` and `codegen-units = 1`. `CEUTrio.HiSeq.WGS.b37.NA12878.20.21.bam`, 654,610 records. Best of three, page cache warm.

| | wall |
|---|---|
| before | 1.458 s |
| after | **1.069 s** |
| `samtools view -h` | 0.97 s |

Profiling before the change put about 20% of the run in malloc and free, with the largest non-allocator items being the boxed quality-score iterator and the per-byte writes; those are what this removes.

## Correctness

Output is byte-identical to what noodles produced before the change, and identical to `samtools view --no-PG -h` on the same input, `cmp` on the whole file. The workspace test suite passes with `--all-features`.

Two commits: the writer change, and the changelog. This is independent of #419 and #423, which are on the read-to-BAM side.

### DIFF
--- a/noodles-sam/CHANGELOG.md
+++ b/noodles-sam/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## Unreleased
+
+### Changed
+
+  * sam/io/writer/record/sequence: Write the sequence in blocks.
+
+    Bases were emitted one at a time, which is a `write_all` call per base of
+    every record. Raw sequences are now written in a single call, and
+    four-bit-packed ones are unpacked into a fixed buffer first. Output is
+    unchanged.
+
 ## 0.90.0 - 2026-08-21
 
 ### Changed
--- a/noodles-sam/src/io/writer/record/sequence.rs
+++ b/noodles-sam/src/io/writer/record/sequence.rs
@@ -3,6 +3,10 @@ use std::io::{self, Write};
 use super::MISSING;
 use crate::alignment::record::{Sequence, SequenceRef, sequence_ref::FourBitPacked};
 
+// Bases are emitted in blocks rather than one at a time: a per-base `write_all` is a call and a
+// bounds check for every base of every record.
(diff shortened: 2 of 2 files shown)


BenjaminDEMAILLE:
Related issue: #424, for the general shape of this. The measurement here was against `samtools view -h` single-threaded; with `-@ 16` samtools does the same conversion in 0.15 s, because it threads the record pipeline and this path does not. That is the same constraint as #424 rather than anything specific to this change, and it does not affect the 1.458 s to 1.069 s the PR reports, which is a like-for-like single-threaded comparison.

